PR #64, opened 2025-09-10
Pattern and destructuring blog post

pataei:
@midnightntwrk/mn-techwriting I'm waiting on technical reviews on this. Once I receive and apply them I'll ask for your reviews

### DIFF
--- a/blog/2025-08-22-pattern.mdx
+++ b/blog/2025-08-22-pattern.mdx
@@ -35,8 +35,8 @@ more readable, and often less prone to errors.
 
 ## What Can You Destructure? 🧩
 
-In Compact, pattern destructuring is primarily designed to be used with values of two key types:
-**tuples** and **structures**. Understanding these types is fundamental
+In Compact, pattern destructuring is primarily designed to be used with values of three key types:
+**tuples**, **vectors**, and **structures**. Understanding these types is fundamental
 to effectively using destructuring.
 
 - Tuple type `[T, ...]`: Tuple types are ordered, heterogeneous
@@ -44,6 +44,10 @@ collections of types (e.g., `[Field, Boolean, Uint<8>]`).
 Destructuring tuple values allows you to bind individual
 elements to identifiers based on their position within a tuple value and typecheck
 the actual element type with the declared type annotation if one exists.
+- Vector type `Vector<n, T>`: Vector types are homogeneous tuple types
+(e.g., `Vector<3, Field>`).
+Destructuring vector values is the same as destructuring tuple values since
+vector values are constructed the same way tuple values are.
 - Structure type `S { f: T ... }`: Structure types are user-defined types with
 named fields (e.g., `struct Point { x: Field, y: Field }`).  Destructuring
 structure values lets you extract values of specific fields by their names and typecheck
@@ -59,7 +63,7 @@ several key areas of Compact programming:
 most frequently for local variable assignments, allowing you to unpack values
 from expressions into new variables.
 
-## How to Destructure Patterns in Compact 🛠️
+## How to Destructure Values in Compact 🛠️
 
 Let's dive into the specifics of how to use tuple and structure patterns.
 The grammar for a pattern `p` is provided in
--- a/blog/2025-08-22-pattern.mdx
+++ b/blog/2025-08-22-pattern.mdx
@@ -1,0 +1,308 @@
+---
+slug: destruction-in-compact
+title: Unpacking - A Guide to Patterns and Destructuring in Compact 📦
+authors: parisa
+tags: [compact]
+image: /img/blog/zkp.jpg
+date: 2025-08-22
+---
+# Unpacking - A Guide to Patterns and Destructuring in Compact 📦
+
+Compact offers powerful features to streamline your contract development.
+One such feature, pattern destruction, allows for elegant and efficient
+extraction of values from complex data structures like tuples and structures.
+If you're coming from languages like TypeScript, you'll find the concept
+familiar.  This post will introduce you to pattern destruction, explain what
+can be destructured, where it's permitted, and show you how to
+leverage it in your Compact code.
+
+Patterns and destructuring were introduced in
+[Compact 0.14.0](#https://docs.midnight.network/relnotes/compact/compact-0-14-0#you-can-use-typescript-compatible-destructuring).
+The code snippets provided in this blog post are written in Compact 0.17.0.
+In code snippets a `demo` circuit is used to provide a context to demonstrate the
+statements.  In such cases, the meat is the statements inside the body of the `demo` circuit.
+
+## What is Pattern Destructuring? ✨
+
+At its core, pattern destruction in Compact is a syntactic convenience that
+allows you to **unpack** values from data structures (specifically tuples and
+structures) and bind them to new identifiers in a single, concise program
+element.  Instead of accessing individual elements or fields one by one by
+indexing into a tuple or a structure,
+destructuring lets you define a pattern that mirrors the shape of the data,
+directly assigning its components to identifiers. This makes your code cleaner,
+more readable, and often less prone to errors.
+
+## What Can You Destructure? 🧩
+
+In Compact, pattern destructuring is primarily designed to be used with values of two key types:
+**tuples** and **structures**. Understanding these types is fundamental
+to effectively using destructuring.
+
+- Tuple type `[T, ...]`: Tuple types are ordered, heterogeneous
+collections of types (e.g., `[Field, Boolean, Uint<8>]`).
+Destructuring tuple values allows you to bind individual
+elements to identifiers based on their position within a tuple value and typecheck
+the actual element type with the declared type annotation if one exists.
+- Structure type `S { f: T ... }`: Structure types are user-defined types with
+named fields (e.g., `struct Point { x: Field, y: Field }`).  Destructuring
+structure values lets you extract values of specific fields by their names and typecheck
+the actual field's type with its declared type annotation if one exists.
+
+## Where Can You Use It? 📍
+
+Pattern destructuring isn't just for a single use case; it's integrated into
+several key areas of Compact programming:
+
+- Parameters of circuits, anonymous circuits, and a constructor can be a pattern.
+- `const` binding statements can bind a pattern. This is arguably where you'll use destructuring
+most frequently for local variable assignments, allowing you to unpack values
+from expressions into new variables.
+
+## How to Destructure Patterns in Compact 🛠️
+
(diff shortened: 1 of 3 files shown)


blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+
+Tuple values can be destructured using a list of patterns enclosed in `[` and `]`.
+You can destructure a tuple into a sequence of identifiers:
+
+```compact
+const [a, b] = [1, true]; // 'a' is 1, 'b' is true
+```
+
+Sometimes you don't need all the values in a tuple. Compact allows you to skip
+elements by simply omitting the identifier but you cannot skip the actual element:
```
kmillikin, line 83: You might say that this behavior is the same as TypeScript (at least, I think it is).  That implicitly addresses a Compact criticism that this is weird and unnecessary (for people coming from languages like ML or Haskell).

I think it's strange to say that you "cannot skip the actual element".  In a sense, the pattern does skip (naming) the actual element.  I think what you should say is that there is no equivalent for constructing tuple values.  You can't leave off an element when constructing a tuple value.

blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+
+```compact
+const [x, , z] = [10, 20, 30]; // 'x' is 10, 'z' is 30, 20 is skipped
+// const [x, , z] = [10, , 30]; // triggers a parse error
+```
+
+Skipped identifiers count towards the length of the pattern if there exists a
```
kmillikin, line 90: This is the introduction of the fact that patterns have a length, but it's not used until much later.  I suggest either introducing it later where it's needed, or elaborating here.

blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+
+Let's dive into the specifics of how to use tuple and structure patterns.
+The grammar for a pattern `p` is provided in
+[Compact's reference](https://docs.midnight.network/develop/reference/compact/lang-ref#parameters-patterns-and-destructuring).
+
+
+### Destructuring Tuple Values
+
+Tuple values can be destructured using a list of patterns enclosed in `[` and `]`.
+You can destructure a tuple into a sequence of identifiers:
```
kmillikin, line 76: Hmmm.  "A sequence of identifiers" sounds a little bit like a single thing that I might be able to manipulate.  `[a, b]` is just syntax, not an actual sequence to me.  I might avoid it and say "You can destructure a tuple by binding its elements to identifiers:"

blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+
+```compact
+const [a, b] = [1, true]; // 'a' is 1, 'b' is true
+```
+
+Sometimes you don't need all the values in a tuple. Compact allows you to skip
```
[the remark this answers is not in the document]
kmillikin: Maybe "Sometimes you don't need to name all the elements in a tuple.  Compact allows you to leave them unnamed by simply omitting the identifier..."

blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+:::note
+If you want to enforce length 'l' of a tuple pattern you must have a named identifier
+at the `l-1` index of your pattern. For example, if you need to enforce a tuple
+pattern of legth 3 you need to have a named identifier on index 2 (e.g., all
+`[, , x]`, `[, x, y]`, and `[x, y, z]` enforce a tuple pattern of length 3).
+:::
+
+### Type Annotations for Tuple Patterns
+
+In a circuit definition, type declarations are required for tuple patterns:
```
kmillikin, line 103: We haven't seen tuple patterns for parameters yet, it might be better to introduce them earlier.

I don't know if you want to spell it out, but it might be useful to give people the model that patterns for parameters are purely a convenience.  Without it they could get exactly the same effect by immediately (or just before they need to) destructuring the parameter:

```
circuit example(tuple0: [Field, Boolean]): Field {
  const [x, y] = tuple0;
  return x;
}
```

As a convenience we allow them to move the pattern into the circuit declaration by replacing the right-hand side identifier in the destructuring `tuple0` with the left-hand side pattern. (That also helps explain why the type is still needed there.)

kmillikin: An alternative, which is a bigger rewrite, is to talk about *everything* in terms of const binding, and then only at the end talk about parameters as a special case of const binding.

blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+}
+```
+
+Such a type declaration dictates the structure and the type of each element.  For example,
+in `exampleTuplePattern` the input has to be a tuple, it first element has to be a value
+of type `Field`, and its second element has to be a value of type `Boolean`.
+A call to `exampleTuplePattern` that does not meet these restrictions
+(e.g., `exampleTuplePattern([1, 1])`) will trigger a type error.
+
+For const bindings and anonymous circuits, type annotations are optional but recommended
```
kmillikin, line 118: I'm not sure I agree with this recommendation.  I would personally favor a style that left those types off, because they are kind of verbose and distracting or because I don't care to work out exactly what they are and have to update type annotations every time I do a benign refactoring.  (Would we make the same recommendation for const binding in general, no patterns?)

blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+for clarity and an easier debugging experience. When a type annotation is provided
+ensure its structure aligns with the pattern. For example, `[a, , b] : [Field, Field, Field]`
+aligns the structure of the pattern and its type whereas `[a, , b] : [Field, Field]` does not:
```
kmillikin, line 121: The second of these is a type error, isn't it?  The sentence reads like this is a style recommendation, but I think (hope) it's a requirement.

blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+the pattern states that it is a tuple with three elements but the type annotation is
+tuple type of length two.
+
+Unlike patterns, types do not allow skipping elements in a tuple type. For example,
```
kmillikin, line 125: This is an interesting one.  What is the motivation?  To me, I feel like I ought to be able to do that (omit a type for an unnamed element, I certainly can't refer to it so why do I care what the type is?  If that type changes on the right-hand side, I really don't want to have to update this type annotation for the value I don't care about.)

In fact, I'd actually like to be able to have type inference for some elements and write:

```
const [x, y, z]: [Field, , Uint<32>] = foo();
```
Where the type annotations are checked and the missing one is inferred from the return type of `foo`.

kmillikin: I feel like restrictions like this are important to justify a little bit from a design standpoint.  Is there a technical reason we need it?  Do we just like it that way?  Is it something that we might change?

blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+`[a, , b] : [Field, , Field]` causes a parse error. So even when you're skipping an
+element in a pattern you must assign it a dummy type. Unfortunately, you have to
+remember what dummy type you used when you're passing arguments to pattern parameters.
```
kmillikin, line 128: I don't think I would call this a "dummy type".  You have to pick a concrete type.

What you might want to do is make this circuit generic:

```
circuit second<T>([, x]: [T, Boolean]): Boolean {
  return x;
}
```
And then you have to specialize it correctly at the call site.

blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+  return haveToUseDummyType([1, true]);
+
+  // return haveToUseDummyType([true, true]);
+  // triggers a type error since the first element must be
+  // a value of type Field
+}
+```
+
+:::note
+We recommend you using a sepecific type as your dummy type throughout all your Compact
```
kmillikin, line 146: I don't understand this completely.  Following the recommendation, the function works to extract the second element of a tuple of type `[Boolean, Boolean]`.  But if I've got a value `t` of some other tuple type `[Field, Boolean]` then I have to do something like:
```
const [, y] = t;
... haveToUseDummyType([false, y]) ...
```

blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+code or use a type synonym for such a dummy type when such a feature is available. For
+example, always use `Boolean` as your dummy type.
+:::
+
+Similar to tuple types, skipped elements of a tuple value cannot be dropped, for example,
+`const [x, , y] : [Field, Field, Field] = [1, , 3];` triggers a type error.
+
+When a type annotation exists, the length of the tuple pattern must be less than or equal to
+the length of its tuple type annotation.
+For example, `[a, b]: [Field, Field]` typechecks.
```
kmillikin, line 156: I probably wouldn't say that this type checks (whether it does or not depends on what is on the left-hand side or what is given as the argument at a call site).  It's just not a syntax error (and the third one below is a syntax error).  It's a structural check.

blog/2025-08-22-pattern.mdx
```diff
@@ -0,0 +1,312 @@
...
+So does `[, a]: [Field, Field]`. So does `[a, , , , ,] : [Field, Field]`.
```
kmillikin, line 157: I *think* this is best explained as: we allow a trailing comma in tuples because that can be convenient.  We treat any number of trailing commas the same as a single one.  So the pattern `[a, , , , ,]` is equivalent to `[a,]` which is equivalent to `[a]`.

kmillikin: And then, after removing the trailing commas, you introduce fresh identifiers for all the missing elements.  So `[, , a]` is equivalent to `[, , a]` (itself) after removing all the (zero) trailing commas.  And that's equivalent to `[_0, _1, a]` provided those names are fresh.

kmillikin (commented):
A whole bunch of comments to take or leave as you wish.

nebelsprachen:
Hi @pataei! Thank you for this PR contribution. What is the status of it? Is it ready for a docs review. Also could you please ensure your commits are signed; CI will block the merge if all commits are not signed

nebelsprachen:
image

@pataei the commits in this PR need to be signed, for this to merge.

nstanford5:
@pataei @kmillikin this is being closed as outdated. If it is still relevant, the PR can be cleaned up and reopened from the closed tab when authors are ready to do so.

PR #185, opened 2025-10-09
Add Preview Network to the Wallet Guide

nicos-shielded:


nebelsprachen:
@nicos-shielded Is the endpoint live? I  think in the chat it was not clear. https://shielded.slack.com/archives/C09JP8ERMK3/p1762778404840159

PR #380, opened 2025-11-17
fix: update pragma directive to use >= 0.16 instead of version range

Olanetsoft:


kmillikin (commented):
I'm not in favor of this change.  I'd just use the current language version, and definitely not a range without an upper bound.

PR #407, opened 2025-11-26
fix: address code review feedback on pragma directive and contract ex…

Olanetsoft:
…amples

- Change from open-ended version (>= 0.16) to exact version (0.17)
- Clarify that pragma is optional but recommended best practice
- Remove unnecessary standard library import (contract doesn't need it)
- Use 2-space indentation (official Compact style)
- Change <function> to <name> in syntax examples to avoid TypeScript keyword confusion
- Use consistent angle bracket syntax in circuit syntax examples
- Clarify that disclose is an operator, not a function


nebelsprachen:
@kmillikin do you want to review this as well?

PR #461, opened 2025-12-10
docs: Add links to Faucet and Releases in Module 5

kwadwoatta:
This PR adds hyperlinks to the 'Midnight Testnet Faucet' and the 'Midnight releases page' (for VS Code extension) in Module 5, improving navigation for developers.

### DIFF
--- a/.github/workflows/apis.yml
+++ b/.github/workflows/apis.yml
@@ -45,7 +45,7 @@ jobs:
 
     steps:
       - name: Checkout docs repository specific branch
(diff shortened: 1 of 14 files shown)


gyrationtechs (changes requested):
@kwadwoatta Your changes are good to go. However, you're missing an important step. Commits need to be signed before creating a PR. 

The good news is that you can resign an already published commit. See our [contributing guide](https://docs.midnight.network/contribute#verified-commits-required) for details on how to set up signing.

PR #478, opened 2025-12-17
fix: remove invalid quotes from docker proof-server command

m4xshen:
Running the proof server docker command with single quotes produces errors. All occurrences are fixed in this PR.


m4xshen:
> Good catch.
> 
> Please sign your commit and resubmit. Thank you

done!

PR #506, opened 2025-12-29
Fix/wallet sdk relnote

gyrationtechs:


gyrationtechs:
Will be resolved by this PR #505

PR #511, opened 2025-12-31
feat: Update site to use SSG instead of client side rendering

gyrationtechs:
Highlight of changes:
- Removed BrowserOnly wrapper from the Layout/index.js file to ensure static content is generated when the site is built.
- Used the IsBrowser component to wrap the Auth0Provider component that requires browser context.
- Removed the rewrites line from the vercel.json file. This line forces vercel to render only the index.html page as opposed to rendering all files.
- The BraveShieldWarning component requires browser context. Added a conditional logic to only execute it in the browser when needed.

Steps to test SSG on the Vercel Preview:
- [ ] Open the preview on your browser.
- [ ] Then, disable JS.
- [ ] Reload the page. 

To disable JavaScript:
- Open Developer Tools then type **CMD + Shift + P** (or CTRL + Shift + P on Windows) to run a command.
- In the command field, type ‘javascript’ and select **Disable JavaScript**.
- Finally, reload the page. 

If content is still showing on the page, SSG is working.

> **Note:** Repeat the same process to enable JavaScript on the preview.

### DIFF
--- a/src/components/BraveShieldsWarning.tsx
+++ b/src/components/BraveShieldsWarning.tsx
@@ -19,7 +19,7 @@ export default function BraveShieldsWarning() {
   const location = useLocation();
 
   useEffect(() => {
-    // Only run in browser environment
+    // Only run in browser environment.
     if (!ExecutionEnvironment.canUseDOM) return;
 
     const checkBrave = async () => {


Olanetsoft (dismissed):
Worked fine for me.

DianaHackmamba (approved):

PR #515, opened 2026-01-02
fixed broken link

Tosin-webdev:
changed link pointing from https://docs.midnight.network/learn/introduction/what-is-midnight
 to https://docs.midnight.network/learn/what-is-midnight


gyrationtechs (approved):
Good catch 🎉

PR #640, opened 2026-02-19
Docs: New doc for the bboard example and smart contract tutorial

gyrationtechs:
This PR:
- Creates a new example doc for the bboard DApp
- Creates a new tutorial doc for the bboard smart contract
- Hides outdated bboard example docs
- Hides outdated bboard contract doc


docs/tutorials/beginner/bboard/smart-contract.mdx
```diff
@@ -0,0 +1,715 @@
...
+export ledger message: Maybe<Opaque<"string">>;
+
+export ledger sequence: Counter;
+
+export ledger owner: Bytes<32>;
+```
+
+Each ledger field serves a specific purpose:
+
+- `state`: Tracks whether the board is vacant or occupied
```
kmillikin: I definitely appreciate that you are not introducing this here, and I think it's totally acceptable to just commit this as a clear improvement and make bigger changes later.

I also appreciate that there is a tension between (a) illustrating language features and (b) using best or recommended practices.

HOWEVER, it has always bothered me that this is redundant.  The state is VACANT exactly when message is none, and state is OCCUPIED exactly when message is some(m).

I would instead adopt one of: (a) keep the state field, just make the message a plain opaque string (not optional), or (b) remove the state field and use a none message to indicate that there is no message on the bulletin board.  A way to think of it is: the Maybe type is tagged with a boolean flag indicating whether it is some or none, and the contract state is tagged with a redundant flag (the two-value enum is just a boolean value).  We don't need both tags.

Maybe<string> is most useful when you need to distinguish between an empty string (the default value) and the absence of a string.  But we don't actually have that here because we already have the state, and we don't allow a person to post "no message" (as opposed to an empty message) because the post circuit takes a string, not a maybe string.

docs/tutorials/beginner/bboard/smart-contract.mdx
```diff
@@ -0,0 +1,715 @@
...
+- **Verify ownership**: Compares the regenerated commitment to the stored `owner` value
+- **Extract message**: Uses `message.value` to access the inner value from the `Maybe` type
+- **Update state**: Changes `state` to `State.VACANT`
+- **Increment sequence**: Advances the counter so the next post uses sequence number 2, 3, and so on
+- **Clear message**: Resets `message` to `none`
+- **Return message**: Returns the taken-down message to the caller
+
+The second `assert` is where privacy meets access control. The user proves they can regenerate the stored commitment without revealing their secret key. The Zero Knowledge (ZK) proof validates this assertion without exposing the private data.
+
+Note that `message` is cleared after extracting its value. This ensures the board is ready for the next post.
```
kmillikin: The board would equally be ready for another post if the message was not cleared (in the current architecture, where there is a state field --- if that was changed to use the message being none, then we would have to clear the message it to indicate the board is vacant).  The post circuit does not assert the message is none.

This is instead a design decision, to clear the board.  You could also have an almost identical contract that didn't do that.  In that case "takeDown" doesn't actually remove the post from the bulletin board, it leaves it up so people can continue to see it, but it just "unlocks" the board so someone can replace that post later.

docs/tutorials/beginner/bboard/smart-contract.mdx
```diff
@@ -0,0 +1,715 @@
...
+// ... previous code ...
+
+export circuit publicKey(sk: Bytes<32>, sequence: Bytes<32>): Bytes<32> {
+  return persistentHash<Vector<3, Bytes<32>>>([pad(32, "bboard:pk:"), sequence, sk]);
+}
+```
+
+This helper circuit:
+
+- Takes a secret key (`sk`) and sequence number as inputs
```
kmillikin: This circuit doesn't know that sk is a secret key or even that it is considered private, it can be anything.

docs/tutorials/beginner/bboard/smart-contract.mdx
```diff
@@ -0,0 +1,715 @@
...
+Note that `message` is cleared after extracting its value. This ensures the board is ready for the next post.
+
+### Create the publicKey helper circuit
+
+The `publicKey` circuit generates a cryptographic commitment to the poster's identity:
+
+```compact
+// ... previous code ...
+
+export circuit publicKey(sk: Bytes<32>, sequence: Bytes<32>): Bytes<32> {
```
kmillikin: I would actually pass the sequence number as the value read from the field (don't read the field here, it makes the circuit impure).  That is, here:

```
export circuit publicKey(sk: Bytes<32>, sequence: Uint<64>): Bytes<32> {
```

and in the body: `sequence as Bytes<32>`.  The call sites don't need the cast, then.

docs/tutorials/beginner/bboard/smart-contract.mdx
```diff
@@ -0,0 +1,715 @@
...
+- Takes a secret key (`sk`) and sequence number as inputs
+- Returns a 32-byte cryptographic hash
+- Uses `persistentHash` from the standard library for secure hashing
+- Creates a vector of three 32-byte values to hash together
+- Uses `pad(32, "bboard:pk:")` to create a domain separator
+
+The commitment has critical properties:
+
+- **Domain separation**: The `"bboard:pk:"` prefix prevents hash collisions with other uses of the secret key
+- **One-way**: You cannot reverse the hash to discover the secret key
```
kmillikin: `persistentHash` is actually SHA-256, which is not reversible.  That's what makes this not reversible.

docs/examples/bboard.mdx
```diff
@@ -0,0 +1,501 @@
...
+│   └── src/test/              # Contract unit tests
+├── api/                       # Application logic shared by interfaces
+│   └── src/                   # Core DApp functionality
+├── bboard-cli/                # Command-line interface
+│   └── src/                   # CLI implementation
+└── bboard-ui/                 # Web browser interface
+    └── src/                   # React-based UI
+```
+
+### The bulletin board contract
```
DianaHackmamba:
```suggestion
## The bulletin board contract
```

docs/tutorials/beginner/bboard/smart-contract.mdx
```diff
@@ -0,0 +1,715 @@
...
+
+```compact
+// ... previous code ...
+
+witness localSecretKey(): Bytes<32>;
+```
+
+This declaration:
+
+- `witness` marks this as a function that accesses private state
```
DianaHackmamba: Complete sentences, add periods.

docs/tutorials/beginner/bboard/smart-contract.mdx
```diff
@@ -0,0 +1,715 @@
...
+}
+
+export circuit publicKey(sk: Bytes<32>, sequence: Bytes<32>): Bytes<32> {
+  return persistentHash<Vector<3, Bytes<32>>>([pad(32, "bboard:pk:"), sequence, sk]);
+}
+```
+
+## Compile the contract
+
+Compilation transforms your Compact code into Zero Knowledge (ZK) circuits and generates TypeScript APIs for interacting with the contract.
```
DianaHackmamba: You only need to define each abbreviation once per page.

docs/tutorials/beginner/bboard/smart-contract.mdx
```diff
@@ -0,0 +1,715 @@
...
+// ... previous code ...
+
+witness localSecretKey(): Bytes<32>;
+```
+
+This declaration:
+
+- `witness` marks this as a function that accesses private state
+- `localSecretKey` returns the user's secret key as a 32-byte value
+- The return value never appears on-chain or in public ledger state
```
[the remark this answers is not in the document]
gyrationtechs: Got it. Updated to use "doesn't"

docs/tutorials/beginner/bboard/smart-contract.mdx
claudebarde: It is indeed redundant, but I always thought that the whole point of the `state` value was to demonstrate how enums work. I believe it is worth keeping it, for learning purposes.

PR #666, opened 2026-02-27
docs: proper hello world

nstanford5:
Major Changes:
- Reduces Hello World from **1,300+ lines of .md to ~300**, combining Compact tutorial with Deploy and Interact guides
- Actually writes Hello World to the blockchain
- Significantly reduces friction and potential failures due to user error
- Makes better use of existing support tooling (npm scripts)
- Deployment Failed! turns a bug into a feature -- running deploy script *always* fails the first time due to slow DUST generation. This section turns it instead into a learning opportunity for users and creates an inline troubleshooting solution.

Minor changes:
- Installation
- deploy-mn-dapp
- bump language_version to 0.21


docs/getting-started/hello-world.mdx
```diff
@@ -74,84 +60,31 @@ Create a new file named `hello-world.compact` in the `contracts` directory:
...
-```
-
-This directive tells the compiler that your contract requires Compact version 0.20.
-
-</Step>
-<Step>
-
-## Define the ledger state
-
-The on-chain state of your contract is declared with `ledger` declarations. On-chain state is public and persistent on the blockchain.
```
laurenelee: I think it's a real risk cutting this line. devs... almost every dev we meet gets confused by this concept.

PR #676, opened 2026-03-02
Changes for compact language reference

pataei:
Some beautifying and structural changes and adding keywords to the sidebar.


pataei:
The current order of the sidebar menu is not based on importance and the order in which a developer will probably need information (see attached screenshot). I'd like to change the order of the menu but I don't know how.  This is the order:

-- Compact
     |-- Writing a contract
     |-- Reference
           |-- Compact reference
           |-- Explicit disclosure
           |-- Formal grammar
           |-- Keywords and reserved words
     |-- Standard library
     |-- Data types
     |-- Compilation and tooling
     |-- Testing and debugging
     |-- Security best practices

PR #680, opened 2026-03-03
Created generating dust guide

JAlbertCode:
Created generating dust guide

### DIFF
--- a/docs/guides/generating-dust-programmatically.mdx
+++ b/docs/guides/generating-dust-programmatically.mdx
@@ -1,29 +1,19 @@
 ---
-title: Generating DUST Programmatically
+title: How to generate DUST programmatically on Preprod
 description: Guide on generating DUST programmatically on the Midnight Preprod environment using the Midnight wallet SDK.
 sidebar_position: 20
-sidebar_label: Generating DUST Programmatically
+sidebar_label: How to generate DUST programmatically on Preprod
 tags: [deployment, Dust, Night, JavaScript]
 toc_max_heading_level: 2
 ---
 
-# How to Generate DUST on Preprod
-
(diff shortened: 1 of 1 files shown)


docs/guides/generating-dust-programmatically.mdx
```diff
@@ -0,0 +1,1036 @@
+---
+title: Generating DUST Programmatically
+description: Guide on generating DUST programmatically on the Midnight Preprod environment using the Midnight wallet SDK.
+sidebar_position: 20
+sidebar_label: Generating DUST Programmatically
+tags: [deployment, Dust, Night, JavaScript]
+toc_max_heading_level: 2
+---
+
+# How to Generate DUST on Preprod
```
gyrationtechs: I think we shouldn't restrict this to Preprod only. Correct me if I'm wrong, but I believe this should also work for `preview` and `undeployed` networks. The user would only need to update the network parameters and config.

We can update it to something like; **How to generate DUST programmatically**

docs/guides/generating-dust-programmatically.mdx
```diff
@@ -0,0 +1,1036 @@
...
+
+## Changes to DUST
+
+DUST is the resource used to process transactions on the Midnight network. In Testnet-02 developers were able to visit the faucet and grab DUST directly from the faucet. This enabled developers to start testing a primitive version of the network but DUST is non-transferable and therefore cannot be sent from one wallet to another. You can read [this article on the implications of NIGHT and DUST for developers](https://midnight.network/blog/the-implications-of-night-and-dust-for-developers) to understand why DUST was designed this way.
+
+On Preprod, developers must gather test NIGHT (tNIGHT) tokens from the faucet, and then designate a wallet to generate DUST in. Here are the step-by-step instructions on generating DUST to begin processing transactions on the network to test applications.
+
+## Generating DUST Programmatically
```
gyrationtechs: If you decide to take out the sections above, we can remove this heading to avoid redundancy

PR #803, opened 2026-04-07
Midnight.js v3.2.0 release notes from product repository

MidnightCI:
Automated release notes copy — component: `midnight-js`, source_url: `https://github.com/midnightntwrk/midnight-js/blob/main/docs/releases/v3.2.0/release-notes.md`, source_path: `/docs/releases/v3.2.0/release-notes.md`, version: `v3.2.0`.


DianaHackmamba (commented):
Leaving suggestions as comments so I don't block this.

Generally we like to add links to documentation updated as a result of the change. A placeholder might be helpful:

XX was updated as a result of this change.

XX, XX, and XX were added as a result of this change.

PR #926, opened 2026-05-19
fix: correct circuit name typo in explicit-disclosure.mdx

Mechack08:
## Summary

Fixes a typo in the explicit disclosure reference page where a code example
used the misspelled circuit name `recordRalance` instead of `recordBalance`.

## Changes

- `docs/compact/reference/explicit-disclosure.mdx`: renamed `recordRalance` → `recordBalance`

## Why this matters

The first code block on the page intentionally shows a *working* example of
`disclose()` usage, but the misspelled name is inconsistent with the corrected
example shown directly below it. A developer following along or copy-pasting
the snippet would get unexpected behaviour from their TypeScript API since the
generated function name would be wrong.

### DIFF
--- a/docs/compact/reference/explicit-disclosure.mdx
+++ b/docs/compact/reference/explicit-disclosure.mdx
@@ -24,7 +24,7 @@ import CompactStandardLibrary;
 witness getBalance(): Bytes<32>;
 export ledger balance: Bytes<32>;
 
-export circuit recordRalance(): [] {
+export circuit recordBalance(): [] {
   balance = disclose(getBalance());
 }
 ```


oduameh:
Hi @Mechack08, thanks for this PR the compact team will look into this typo in the main source file.

kmillikin:
Hi @Mechack08.  Compact documentation is maintained by the LFDT Minokawa project.  This file is merely a copy of the canonical source, and it's overwritten on every Compact release.

You really want to send us a PR here: https://github.com/LFDT-Minokawa/compact/blob/main/doc/explicit-disclosure.mdx.  Note that Linux Foundation does not use CLA, they use DCO (https://wiki.linuxfoundation.org/dco) so you'll also have to sign off on your commits (for instance, commit with `commit -s` in git; note that "signing off" is different than "signing" your commits).

PR #958, opened 2026-05-26
Add concept page: Compact as a privacy-first language

oduameh:


docs/concepts/how-midnight-works/compact-privacy-first-language.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+## Public, private, and proof environments
+
+Once privacy and proofs become central, the language needs to make different kinds of data and execution visible to the developer.
+
+Compact does this by separating the main environments involved in a Midnight smart contract interaction. In the language, these map to three constructs:
+
+- **`ledger`**: public state that belongs to the on-chain contract state.
+- **`circuit`**: logic that participates in proof generation and encodes the rules validators check.
+- **`witness`**: private data known only to the party executing the interaction, provided by callbacks on the user's machine.
```
nstanford5: Witnesses are not exclusive to retrieving private data -- they are primarily for "off-chain compute". Witnesses can be used for sizeable calculations that aren't economically feasible to store on-chain, to enable functionality that doesn't exist in Compact, or they can be used to retrieve data from private state, but that same private state data can be passed as a circuit parameter and still maintain privacy.

For example -- Compact doesn't support the math operation for division, but you can define a witness in Typescript that does this and verify proper execution in-circuit. This pattern is best demonstrated here: https://docs.midnight.network/examples/contracts/calculator -- note that the `divide` circuit doesn't actually divide, it off-loads this work to a witness and uses `assert` to enforce the correct logic in `witness divMod`.

This can be referred to as "on-chain verification of off-chain compute". 

Historically, we have used `witnesses` for simple retrieval of private data -- but they are much more powerful than that.

docs/concepts/how-midnight-works/compact-privacy-first-language.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+For a walkthrough of how these constructs work together in practice, see [Writing a contract](/compact/reference/writing).
+
+This separation maps back to the trust model. The public side needs to be visible to the network. Developers should not reveal the private side unnecessarily. The proof connects the two by showing that the hidden inputs and local computation justify the public result.
+
+Without explicit boundaries, developers lose track of where a value lives, who can see it, and what they must prove about it. That risk is especially high in privacy-preserving applications.
+
+A value meant to stay private should not become public by accident. The network should not accept a public state transition without proof that the interaction satisfied private-side rules. Developers should be able to see, from the shape of the smart contract, where those boundaries are.
```
nstanford5: Maybe avoid the word "proof" unless you are referring to a zk-proof. Maybe "without verification"

docs/concepts/how-midnight-works/compact-privacy-first-language.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+## Explicit disclosure turns privacy into a deliberate choice
+
+Privacy-preserving systems still need disclosure.
+
+A payment system may need to reveal that a transaction is valid. A compliance workflow may need to show a specific fact to an authorized party. A game may need to update a public score while keeping some player input private.
+
+The question is not whether anything should ever become public, but who sees what, and when.
+
+Compact enforces explicit disclosure through the [`disclose()` keyword](/compact/reference/explicit-disclosure). The compiler rejects any program that writes witness-derived data to the ledger, returns it from an exported circuit, or passes it to another contract without a `disclose()` wrapper. Privacy is the default; disclosure requires a deliberate declaration.
```
nstanford5: There is a lot of misunderstanding out there about the `disclose` keyword. The keyword itself does not make anything public. It tells the compiler "the developer is marking this value as safe to store publicly" -- values only become public when they cross into a public domain (ledger writes, returns from exported circuits, c2c calls). This paragraph doesn't necessarily get this wrong, but could be more clear about the `disclose` keyword.

docs/concepts/how-midnight-works/compact-privacy-first-language.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+If you prefer to set up your environment first, start with [Install the toolchain](/getting-started/installation) and [Create your first contract](/getting-started/hello-world).
+
+## Additional resources
+
+- [Compact language](/compact): overview of Compact and its role in generating zero-knowledge circuits.
+- [Smart contracts on Midnight](/concepts/how-midnight-works/smart-contracts): the transcript, proof, and verifier key model that Compact compiles into.
+- [Zero-knowledge proofs](/concepts/zero-knowledge-proofs): how ZK Snarks work and how Midnight uses them.
+- [Kachina](/concepts/kachina): the formal model behind public and private state separation.
+- [Example contracts](/examples/contracts/calculator): browse Compact contracts for elections, auctions, token transfers, and more.
```
nstanford5: I think this path is intending to stop at the contracts category -- not pointing at the calculator contract itself

docs/concepts/how-midnight-works/compact-privacy-first-language.mdx
oduameh: Thanks for the correction. I've updated the witness description to frame them as off-chain functions for compute, not just private data retrieval. Added a reference to the calculator example to show the division/witness verification pattern you mentioned.

docs/concepts/how-midnight-works/compact-privacy-first-language.mdx
oduameh: Good call. Changed to "verification" to keep "proof" reserved for ZK proofs throughout the page.

docs/concepts/how-midnight-works/compact-privacy-first-language.mdx
oduameh: Appreciate the nuance here. I've rewritten this paragraph to make the distinction clear: disclose() is a compiler annotation that marks a value as safe to enter a public domain. The value only becomes public when it crosses a visibility boundary (ledger write, exported circuit return, or contract-to-contract call). The compiler enforces that no witness-derived value crosses those boundaries without the annotation.

docs/concepts/how-midnight-works/compact-privacy-first-language.mdx
oduameh: Fixed the link to point to /category/contracts so it lands on the category page, not the calculator specifically.

docs/concepts/how-midnight-works/compact-privacy-first-language.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+## Public, private, and proof environments
+
+Once privacy and proofs become central, the language needs to make different kinds of data and execution visible to the developer.
+
+Compact does this by separating the main environments involved in a Midnight smart contract interaction. In the language, these map to three constructs:
+
+- **`ledger`**: public state that belongs to the on-chain contract state.
+- **`circuit`**: logic that participates in proof generation and encodes the rules validators check.
+- **`witness`**: off-chain functions that run on the user's machine. You can use witnesses to retrieve private state, but also to perform computation that is not feasible or not supported on-chain, such as division. The contract verifies witness results in-circuit, giving you on-chain verification of off-chain compute. See the [calculator example](/examples/contracts/calculator) for a practical demonstration of this pattern.
```
nstanford5:
```suggestion
- **`witness`**: off-chain functions that run on the user's machine. You can use witnesses to retrieve private state, but also to perform computation that is not feasible or not supported on-chain, such as division. The contract verifies witness results in-circuit, giving you on-chain verification of off-chain compute. See the [calculator example](/examples/contracts/calculator) for a practical demonstration of this pattern. Witnesses are also critical for updating private state based on in-circuit computation.
```

nstanford5 (changes requested):

nstanford5 (approved):

PR #980, opened 2026-06-01
docs: document non-ledger transactions and classification

m2ux:
## Summary

Augments the network-architecture transactions page to document the Midnight transaction format, its classification vocabulary, and the node-owned non-ledger transaction categories. This is the docs-site half of the work for midnightntwrk/midnight-node#1533; the rustdoc half lands in midnightntwrk/midnight-node#1619.

## What changed

- Adds a transaction-format and classification spine, framing `TransactionTypeV2` as an off-node vocabulary consumed through runtime metadata, not an in-node dispatch mechanism.
- States the ledger/non-ledger ownership boundary explicitly and defers ledger payload internals to the ledger team via an `:::info` admonition linking the ledgers concept page and the ledger API reference.
- Adds a non-ledger category reference table plus per-category sections: system transaction, timestamp, unknown, cNIGHT observation, and bridge transfer. The bridge transfer is presented as an inherent that produces a system transaction, never as a peer extrinsic.
- Mentions the partner-chains consensus inherents (AURA slot, main-chain hash, authority selection) with pointers to the onchain-logic and consensus pages, without giving them per-category entries.
- Preserves the existing ledger transaction lifecycle prose.

## Verification

Docusaurus build, Vale, markdown link-check, and Prettier run as the repo's CI gates.

Assisted by AI


m2ux:
I went through the eleven Vale alerts and pushed a follow-up commit (0bcee43c).

The spelling alerts on inherent, inherents, and onchain are legitimate domain vocabulary rather than typos, so I added those terms to the Midnight accept-list, which is how this repo already handles words like extrinsic and Substrate. Worth noting the repo config disables the spelling style outright, so these alerts came from the action's default rules rather than house policy. Onchain in particular is already in active use as the title of the sibling onchain-logic page.

I also took the active-voice suggestion on the ledger-ownership sentence and reworded it, which reads better with no loss of meaning.

I left the rest as-is. The runtime possessive is just standard English that the default speller flagged on the apostrophe. The repetition alert does not correspond to an actual repeated word on that line; it is the experimental alpha rule misfiring. And the second-person and remaining active-voice alerts both land on a sentence where the passive is doing real work: the user transaction pool is a compound noun rather than the reader being addressed, and describing what governance is permitted to run is an accurate way to state a system-imposed gating constraint that an active rewrite would distort.

*Posted by an agent on behalf of Mike Clay.*

nstanford5 (changes requested):
@m2ux commits need verified signatures

oduameh:
This work has been taken over in #1131. Thank you for the original contribution, @m2ux!

PR #983, opened 2026-06-02
Add card descriptions to Examples and Tutorials subcategories

oduameh:


nstanford5 (changes requested):
Instead of moving the descriptions up a level, we should be adding some net new summaries of the value of each tutorial category. There may be some limitations of the Docusaurus "category cards" -- but we need to discover what else is possible here.

For example, the Battleship tutorial describes its value at the beginning of the smart contract tutorial (https://docs.midnight.network/tutorials/bship/smart-contract) -- we need a summary similar to these at the top level of each category so that developers know what content is best for their intentions at a glance

PR #1018, opened 2026-06-09
Add Midnight Indexer v4.3.3 release notes

oduameh:


nstanford5 (dismissed):

mzf11125 (commented):
Good addition overall. A few observations after reading through the release notes and the component update:

**`midnight-indexer-4-3-3.mdx`**

The frontmatter and structure follow the existing release note conventions correctly. The caution callout for the 4.0.x to 4.3.3 upgrade path with the 3-4 day re-index estimate is really helpful for operators.

One thing worth checking: the breaking change section says the rename is confined to `@beta` fields. The note is accurate but it might be worth adding a one-liner explicitly stating that consumers on the stable 4.0.x API surface are completely unaffected, since operators on preprod/mainnet who have not yet upgraded to Node 1.0.0 may read this and worry unnecessarily.

Also the `dustGenerations` bug fix description is clear but the linked PR number (#1166) could optionally include a short inline description of the root cause (the dtime drain short-circuit) so someone can understand the fix without following the link.

**`DynamicListMidnightIndexer.js`**

The previous LATEST entry (4.0.2) was correctly demoted to SUPPORTED. One suggestion: the 4.0.x line is still the production line for Node 0.22.x operators. It might be worth adding a `note` field on the 4.0.x entry like `'Maintenance line for Node 0.22.x'` so the dynamic list communicates which line to track based on node version, matching what the release notes prose already says.

Overall the content is accurate and well-structured. Nice work.

nstanford5 (dismissed):

PR #1045, opened 2026-06-13
docs(compact): add "Disclosing low-entropy values safely" to explicit-disclosure

CjDabrow:
## What
Adds a new section, "Disclosing low-entropy values safely", to `docs/compact/reference/explicit-disclosure.mdx`, right after the existing "Safe Compact standard library routines" section.

## Why
The page explains that the compiler treats `transientCommit` and `persistentCommit` as safe to disclose, but it does not warn about a common and subtle pitfall: disclosing a hash, or a commitment built with weak randomness, of a low-entropy value does not keep that value secret. An observer can brute-force the small input space. On a privacy chain this is exactly the kind of mistake that leaks "private" data, so it is worth calling out where developers learn about disclosure.

## What the section adds
- A plain explanation that `disclose()`, and the compiler's "safe" treatment of commitments, controls when disclosure is permitted, not whether the value stays secret in practice.
- A worked unsafe example (disclosing `persistentHash` of a yes/no vote) and the safe alternative (`persistentCommit` with a fresh secret salt).
- A note that the compiler does not verify the commitment's randomness is secret, fresh, or unique, so a reused or predictable salt re-exposes the value.

All Compact signatures used (`persistentHash<T>(value): Bytes<32>`, `persistentCommit<T>(value, rand: Bytes<32>): Bytes<32>`) match the standard library reference.

### DIFF
--- a/docs/compact/reference/explicit-disclosure.mdx
+++ b/docs/compact/reference/explicit-disclosure.mdx
@@ -151,6 +151,39 @@ The message aids the programmer by noting the indirect nature of the disclosure.
 
 The compiler recognizes that certain Compact standard library routines sufficiently disguise witness data so that explicit declaration of disclosure is not required. For an expression `e` whose value contains witness data, the compiler will treat `transientCommit(e)` as if it does not contain witness data, while it will treat `transientHash(e)` as if it does.
 
+## Disclosing low-entropy values safely
+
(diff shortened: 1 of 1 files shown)


dybvig:
Thank you for your PR.  FYI, the [original](https://github.com/LFDT-Minokawa/compact/blob/main/doc/explicit-disclosure.mdx) of this document is located in the [LFDT compact repo](https://github.com/LFDT-Minokawa/compact), and issues or PRs should be posted there instead.

PR #1048, opened 2026-06-15
Fix ~120 Compact-touching 404 URLs via vercel.json redirects

oduameh:
## Summary

Adds 26 redirect rules to `vercel.json` covering ~120 URLs that currently 404 and either start with `/compact/...` (source side) or point at destinations inside `/docs/compact/` (destination side).

Companion to #1047 which covered the non-Compact bulk.

The largest rule is the single wildcard `/compact/reference/midnight-api/:path*` → `/api-reference/:path*`, which alone covers ~100 of the 120 URLs.

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -605,6 +605,137 @@
         "source": "/wallet-api/README.md",
         "destination": "/sdks/official/wallet-developer-guide",
         "permanent": true
+      },
+
+      {
+        "source": "/compact/reference/midnight-api/:path*",
+        "destination": "/api-reference/:path*",
+        "permanent": true
+      },
+      {
+        "source": "/develop/reference/compact/compact-grammar",
+        "destination": "/compact/reference/compact-grammar",
+        "permanent": true
+      },
+      {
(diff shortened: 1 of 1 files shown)


oduameh:
Hi @pasevin, Most of these redirects move users away from old `/compact/reference/midnight-api/...` paths to the new `/api-reference/...` paths. A smaller subset has destinations inside `/docs/compact/`:

- `/compact`
- `/compact/data-types`
- `/compact/reference`
- `/compact/reference/ledger-adt`
- `/compact/reference/compact-grammar`
- `/compact/reference/explicit-disclosure`
- `/compact/standard-library`
- `/compact/compilation-and-tooling`
- `/compact/reference/writing`

The redirects themselves live in `vercel.json` at the repo root, which is owned by the docs team. They will not be touched by your next auto-publish. But if you rename or remove any of the destination pages in `/docs/compact/`, the redirects will start pointing at 404s.

Could you scan and let us know if any of those pages are slated to move or be removed in your next release? If yes, we will either update the destination or hold the relevant rule.

nstanford5 (dismissed):

PR #1073, opened 2026-06-19
docs(troubleshoot): document GitHub Package Registry auth for @midnight-ntwrk/* (closes #787)

### LINKED ISSUE #787: [GH issue][Documentation] Invalid authentication (as an unknown user) installing @midnight-ntwrk packages
### Context & versions

Externally reported issue in `midnight-js`: https://github.com/midnightntwrk/midnight-js/issues/173

### Steps to reproduce

Run `yarn install` (without an `npmAuthToken` configured) on a project that depends on `@midnight-ntwrk/*` packages.

Example `.yarnrc.yml` (without npmAuthToken):
```yaml
compressionLevel: mixed
enableGlobalCache: false
nodeLinker: node-modules
npmScopes:
  midnight-ntwrk:
    npmAlwaysAuth: true
    npmPublishRegistry: "https://npm.pkg.github.com/"
    npmRegistryServer: "https://npm.pkg.github.com/"
yarnPath: .yarn/releases/yarn-4.10.3.cjs
```

### Actual behavior

```
@midnight-ntwrk/ledger@npm:4.0.0::__archiveUrl=...: Invalid authentication (as an unknown user)
```

All `@midnight-ntwrk/*` packages fail to install with an authentication error.

### Expected behavior

Documentation should clearly explain how to configure `npmAuthToken` for the GitHub Package Registry so that users can install `@midnight-ntwrk/*` packages successfully. The error message and resolution path should be covered in the getting started / installation docs.

### Conversation summary

Not specified in original ticket.

---
_Migrated from [PM-19816](https://shielded.atlassian.net/browse/PM-19816)_

[PM-19816]: https://shielded.atlassian.net/browse/PM-19816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

HarleysCodes:
## Summary

Resolves midnightntwrk/midnight-docs#787 by adding a new 'Solution 4' to `docs/troubleshoot/fix-package-repository-access-failures.mdx` covering the GitHub Package Registry authentication requirement that breaks `yarn install` for `@midnight-ntwrk/*` packages with 'Invalid authentication (as an unknown user)'.

## Background

`midnightntwrk/midnight-js`'s own `.yarnrc.yml` configures the `midnight-ntwrk` scope to fetch from `https://npm.pkg.github.com/`. The existing troubleshoot page covers 403s from the public npm registry but never mentions GitHub Packages, so users following its solutions still hit the auth failure.

The canonical fix is already documented in midnightntwrk/midnight-js#173 (adamreynolds-io): add `npmAuthToken: "${GITHUB_TOKEN}"` under the `midnight-ntwrk` scope and set the `GITHUB_TOKEN` env var before installing.

## Change

- Adds Solution 4 between Solution 3 and the existing 'Platform-specific fixes' section.
- Explains why GitHub Packages requires auth even for public packages.
- Documents PAT generation (`read:packages` scope, fine-grained or classic).
- Provides a copy-pasteable `.yarnrc.yml` snippet mirroring the user's working config from the issue.
- Notes that yarn interpolates `${GITHUB_TOKEN}` at install time, so the token never lives in the config file.
- Adds an optional 'inline env var' alternative for users who don't want `export`.

## Verification

- Doc-only change. Build and runtime are unaffected.
- Verified against midnightntwrk/midnight-js's actual `.yarnrc.yml` on `main`: the scope block matches what this PR recommends users add `npmAuthToken` to.
- Cross-referenced the fix against the Yarn 4 `.yarnrc.yml` docs (`npmAuthToken`, `npmAlwaysAuth`, `npmRegistryServer` fields all behave as documented).
- No other doc files reference `npmAuthToken` or `GITHUB_TOKEN`; the change is scoped to the troubleshoot page.

## Out of scope (not changed)

- The install/quickstart guides do not currently mention `@midnight-ntwrk/*` requiring GitHub Packages auth at all. Happy to follow up with a parallel update to those surfaces if the docs team prefers the guidance live there instead of (or in addition to) the troubleshoot page.


docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,76 @@ npm install @midnight-ntwrk/compact-runtime
 
 </StepsProvider>
 
+## Solution 4: Authenticate to GitHub Package Registry
+
+If the previous solutions do not resolve the issue, the failure may be related to GitHub Package Registry authentication rather than your npm registry configuration. Midnight publishes `@midnight-ntwrk/*` packages to `https://npm.pkg.github.com/`, which requires authentication even for public packages.
```
oduameh:
```suggestion
If the previous solutions do not resolve the issue, the cause may be GitHub Package Registry authentication rather than an npm registry misconfiguration. Midnight publishes `@midnight-ntwrk/*` packages to `https://npm.pkg.github.com/`, which requires authentication even for public packages.
```

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,76 @@ npm install @midnight-ntwrk/compact-runtime
...
+
+Without a configured token, every install attempt fails with:
+
+```
+@midnight-ntwrk/ledger@npm:4.0.0::__archiveUrl=...: Invalid authentication (as an unknown user)
+```
+
+To resolve this, configure `npmAuthToken` in your `.yarnrc.yml` and provide a GitHub personal access token with the `read:packages` scope through the `GITHUB_TOKEN` environment variable.
```
oduameh:
```suggestion
To resolve this, configure `npmAuthToken` in your `.yarnrc.yml`. You also need a GitHub personal access token with the `read:packages` scope, provided through the `GITHUB_TOKEN` environment variable.
```

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,76 @@ npm install @midnight-ntwrk/compact-runtime
...
+
+<StepsProvider>
+
+<Step>
+
+### Generate a GitHub personal access token
+
+Create a fine-grained personal access token at `https://github.com/settings/personal-access-tokens/new` with the `read:packages` scope. A classic PAT with the same scope also works.
```
oduameh:
```suggestion
Create a fine-grained personal access token (PAT) at `https://github.com/settings/personal-access-tokens/new` with the `read:packages` scope. A classic PAT with the same scope also works.
```

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,76 @@ npm install @midnight-ntwrk/compact-runtime
...
+
+For organization-owned packages, your token must also have access to the `midnightntwrk` organization.
```
oduameh, line 220:
```suggestion
:::info[Organization access]
For organization-owned packages, your token must also have access to the `midnightntwrk` organization.
:::
```

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,76 @@ npm install @midnight-ntwrk/compact-runtime
...
+```yaml
+npmScopes:
+  midnight-ntwrk:
+    npmAlwaysAuth: true
+    npmPublishRegistry: "https://npm.pkg.github.com/"
+    npmRegistryServer: "https://npm.pkg.github.com/"
+    npmAuthToken: "${GITHUB_TOKEN}"
+```
+
+Yarn interpolates `${GITHUB_TOKEN}` from the environment at install time, so the token never has to live in `.yarnrc.yml`.
```
oduameh:
```suggestion
:::tip[Token security]
Yarn interpolates `${GITHUB_TOKEN}` from the environment at install time, so the token never lives in `.yarnrc.yml`.
:::
```

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,76 @@ npm install @midnight-ntwrk/compact-runtime
...
+
+<Step>
+
+### Test installation
+
+```bash
+yarn install
+```
+
+**Verification**: Installation completes without `Invalid authentication` errors and `@midnight-ntwrk/*` packages resolve from `npm.pkg.github.com`.
```
oduameh, line 267:
```suggestion
**Verification**: Installation completes without `Invalid authentication` errors and `@midnight-ntwrk/*` packages resolve from `npm.pkg.github.com`.
```

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,76 @@ npm install @midnight-ntwrk/compact-runtime
 
 </StepsProvider>
 
+## Solution 4: Authenticate to GitHub Package Registry
+
+If the previous solutions do not resolve the issue, the failure may be related to GitHub Package Registry authentication rather than your npm registry configuration. Midnight publishes `@midnight-ntwrk/*` packages to `https://npm.pkg.github.com/`, which requires authentication even for public packages.
+
+Without a configured token, every install attempt fails with:
+
+```
```
oduameh: For consistency with the error block earlier in the page (which uses ` ```bash `), consider adding a language tag here. Even for output-only blocks, a tag like ` ```text ` or ` ```bash ` helps screen readers and syntax highlighters.

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,76 @@ npm install @midnight-ntwrk/compact-runtime
...
+    npmAuthToken: "${GITHUB_TOKEN}"
+```
+
+Yarn interpolates `${GITHUB_TOKEN}` from the environment at install time, so the token never has to live in `.yarnrc.yml`.
+
+</Step>
+
+<Step>
+
+### Export the token and reinstall
```
oduameh: In Solutions 1-3, the step before "Test installation" only *configures* things, and "Test installation" is the first time `install` actually runs. Here, step 3 both exports the token *and* runs `yarn install`, so step 4 re-runs the same command.

To match the existing pattern, you could separate the concerns:
- **Step 3 ("Export the token")**: just the `export` command plus the inline-env-var alternative
- **Step 4 ("Test installation")**: runs `yarn install` as the actual verification

That way each step does one thing, and the test step is meaningful on its own.

oduameh (commented):
Really well-written addition -- the structure matches the existing solutions cleanly and the technical content is accurate. Just a few suggestions to tighten the prose and align with patterns used in Solutions 1-3.

One item I could not leave inline because it is outside the diff: the page description on **line 3** currently says *"three proven solutions"* but with this PR there are now four. It would be worth updating that to *"four proven solutions"* (or rewording to avoid a specific count).

HarleysCodes:
addressed all 8 inline comments + the line-3 description fix. opted for ` ```text ` on the line 205 error block (output-only, not a command) and folded step 4 into step 3 (since step 3 already runs install — re-running the same command in a separate step added noise). happy to revisit either if you wanted the other shape.

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: Fix package repository access failures
-description: Resolve 403 Forbidden errors when installing Midnight npm packages with three proven solutions.
+description: Resolve 403 Forbidden errors when installing Midnight npm packages with four proven solutions.
```
oduameh: The style guide targets 120-160 characters for descriptions. At 92 characters this falls short. Rewording also avoids a hardcoded count that goes stale whenever a solution is added or removed.

```suggestion
description: Resolve 403 Forbidden and authentication errors when installing @midnight-ntwrk npm packages. Covers registry, network, configuration, and token fixes.
```

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,70 @@ npm install @midnight-ntwrk/compact-runtime
 
 </StepsProvider>
 
+## Solution 4: Authenticate to GitHub Package Registry
+
+If the previous solutions do not resolve the issue, the cause may be GitHub Package Registry authentication rather than an npm registry misconfiguration. Midnight publishes `@midnight-ntwrk/*` packages to `https://npm.pkg.github.com/`, which requires authentication even for public packages.
```
oduameh: Solutions 1-3 use npm. This solution switches to Yarn Berry configuration without signaling the change. One additional sentence helps readers know when this section applies.

```suggestion
If the previous solutions do not resolve the issue, the cause may be GitHub Package Registry authentication rather than an npm registry misconfiguration. Midnight publishes `@midnight-ntwrk/*` packages to `https://npm.pkg.github.com/`, which requires authentication even for public packages. This solution applies to Yarn Berry (v2+) projects that use a `.yarnrc.yml` file.
```

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,70 @@ npm install @midnight-ntwrk/compact-runtime
...
+
+To resolve this, configure `npmAuthToken` in your `.yarnrc.yml`. You also need a GitHub personal access token with the `read:packages` scope, provided through the `GITHUB_TOKEN` environment variable.
+
+<StepsProvider>
+
+<Step>
+
+### Generate a GitHub personal access token
+
+Create a fine-grained personal access token (PAT) at `https://github.com/settings/personal-access-tokens/new` with the `read:packages` scope. A classic PAT with the same scope also works.
```
oduameh: `read:packages` is a classic-token scope. The URL here (`personal-access-tokens/new`) opens the fine-grained token page, where the equivalent permission is called **Packages: Read**. Leading with classic tokens and noting the fine-grained alternative avoids confusion.

```suggestion
Create a classic personal access token at `https://github.com/settings/tokens/new` with the `read:packages` scope. A fine-grained token with the **Packages: Read** permission also works.
```

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,70 @@ npm install @midnight-ntwrk/compact-runtime
...
+### Update `.yarnrc.yml`
+
+Add the `npmAuthToken` field under the `midnight-ntwrk` scope, pointing at the `GITHUB_TOKEN` environment variable:
+
+```yaml
+npmScopes:
+  midnight-ntwrk:
+    npmAlwaysAuth: true
+    npmPublishRegistry: "https://npm.pkg.github.com/"
+    npmRegistryServer: "https://npm.pkg.github.com/"
```
oduameh, line 235: `npmPublishRegistry` is only relevant when publishing packages to the registry. Since this page is about resolving install failures, removing it keeps the YAML focused on the reader's problem.

```suggestion
    npmAlwaysAuth: true
    npmRegistryServer: "https://npm.pkg.github.com/"
```

docs/troubleshoot/fix-package-repository-access-failures.mdx
```diff
@@ -196,6 +196,70 @@ npm install @midnight-ntwrk/compact-runtime
...
+
+:::tip[Token security]
+Yarn interpolates `${GITHUB_TOKEN}` from the environment at install time, so the token never lives in `.yarnrc.yml`.
+:::
+
+</Step>
+
+<Step>
+
+### Export the token and reinstall
```
oduameh: In Solutions 1-3, the step before "Test installation" only configures things, and "Test installation" runs `install` on its own. Here, step 3 bundles export and install in one step. Renaming the heading is the first part of the split. Consider also extracting `yarn install` into a separate Step 4 ("Test installation") with the **Verification** line, matching the pattern in Solutions 1-3.

```suggestion
### Export the token
```

oduameh (changes requested):
Thanks for the addition, @HarleysCodes. This fills a real gap in the troubleshooting coverage. A few changes are needed before merge, mostly around technical accuracy and structural consistency with Solutions 1-3.

### Frontmatter
- The description is 92 characters; the style guide targets 120-160. The suggestion below also drops the hardcoded count so future edits do not require a frontmatter update.

### Context for readers
- Solutions 1-3 use npm commands. Solution 4 switches to Yarn Berry (`.yarnrc.yml`) without flagging the change. Adding one sentence up front avoids confusion for readers who are not using Yarn.

### Technical accuracy
- The `read:packages` scope belongs to classic personal access tokens. The URL in the text (`personal-access-tokens/new`) opens the fine-grained token page, where the equivalent permission is called **Packages: Read**. The suggestion leads with classic tokens and notes the fine-grained alternative.
- `npmPublishRegistry` is only relevant when publishing packages. Removing it keeps the configuration focused on the install problem this page solves.

### Structural consistency
- In Solutions 1-3, each "Test installation" step runs `install` as its own step. Solution 4 currently bundles `export` and `yarn install` in one step. The heading suggestion below is the first half of the fix. Consider also extracting `yarn install` into a new **Step 4: Test installation** so every solution follows the same configure-then-test pattern. (The full restructuring includes fenced code blocks, so GitHub suggestion syntax cannot express it in a single committable block.)

HarleysCodes (commented):
Thanks for the careful pass — the structural-consistency call on Solution 4 is right, I'll extract the install step and add the Yarn-Berry note up front. One thing I'd flag separately: the Vercel preview check is failing on this branch (`license/cla` and the build are green, but the Vercel status check returns failure with an auth/authorize URL), so I can't see a fresh preview of the suggested changes before re-pushing. Looks like a team-config / install-permission issue on the docs site's Vercel side rather than anything content-side. If someone on the docs team can take a look at the Vercel auth for that check, I'll address your review comments in one go and we'll be in good shape.

HarleysCodes:
Addressed the review feedback on PR #1073 in commit 5fdf3618. Summary of the changes (full diff in the Files tab):

- **Frontmatter description** — dropped the hardcoded "four proven solutions" count and rewrote to 144 chars (within the 120-160 style-guide target). Avoids going stale on future edits.
- **Solution 4 intro** — added "This solution applies to Yarn Berry (v2+) projects that use a `.yarnrc.yml` file" so npm-only readers know the section applies to a different stack.
- **PAT generation** — leads with classic tokens (`/settings/tokens/new` with `read:packages`) since `read:packages` is a classic-token scope; the fine-grained alternative is now labelled correctly as **Packages: Read**.
- **`.yarnrc.yml`** — dropped `npmPublishRegistry`. It's only relevant when publishing, not when installing, so it was noise on an install-focused page.
- **Step split** — split "Export the token and reinstall" into two steps to match Solutions 1-3's configure-then-test pattern:
  - **Step 3: Export the token** — just the `export` (with the inline-env-var alternative).
  - **Step 4: Test installation** — runs `yarn install @midnight-ntwrk/compact-runtime`, with the verification line.

All five inline suggestions applied; the structural split you noted couldn't be expressed as a single GitHub suggestion is now in the diff. Ready for another pass whenever you are.

oduameh (approved):
@HarleysCodes — could you either sign the CLA under the `harleycodes` identity, or squash/rebase these commits under your `HarleysCodes` account so the whole PR resolves to one signed identity? That's the only thing standing between this and merge.

PR #1085, opened 2026-06-22
chore: bboard tutorial updates

nstanford5:
- update versions, fixes and imports
- match code here: https://github.com/midnightntwrk/example-bboard/pull/1474


nstanford5:
All of the links here pass

PR #1086, opened 2026-06-22
Fix error in StepsProvider

oluwatobiss:
Added the missing opening code block backticks that caused the steps and code snippet to render inconsistently.


docs/guides/deploy-mn-app.mdx
```diff
@@ -114,6 +114,7 @@ The script:
 
 A successful run ends with output similar to:
 
+```shell
```
oduameh:
```suggestion
```bash
```

oduameh (commented):
Good catch! the missing opening fence breaks the rendering of this output block. One small tweak: every other code block on this page uses `bash` as the language tag, so this one should match for consistency.

PR #1101, opened 2026-06-25
docs(tokens): top-level Tokens category with shielded/unshielded token tutorials

Olanetsoft:
## Summary

Turns **Tokens** into a top-level, expandable sidebar category and adds two tutorials for creating and transferring custom tokens in Compact.

### Changes
- **Category restructure** — `docs/tokens/_category_.yaml` now uses `link: type: generated-index` (same pattern as Guides), so Tokens renders as a real expandable category with **Overview** as its explicit first entry.
- **Overview** — the existing tokens content moved from `index.mdx` → `overview.mdx` (`sidebar_position: 1`), body unchanged aside from a pointer to the two new tutorials.
- **Create and transfer an unshielded token** (`unshielded-token.mdx`, position 2) — transparent token: on-chain balances, `mintUnshieldedToken` / `sendUnshielded`, full deploy-and-test harness.
- **Create and transfer a shielded token** (`shielded-token.mdx`, position 3) — privacy-preserving token: value in immutable coins, witnesses, `mintShieldedToken` / `sendImmediateShielded`, confirmed via the wallet's shielded balance.


oduameh (commented):
(Review withdrawn)

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+
+On Midnight, the privacy model you choose for a token changes how you write its smart contract. This tutorial shows you how to create and transfer both **unshielded** (transparent) and **shielded** (privacy-preserving) tokens in Compact.
+
+You build two tokens in turn. The unshielded token keeps its balances on-chain, where anyone can read them. The shielded token hides its balances, which forces a different design. Putting them side by side is the fastest way to understand how privacy refines a token on Midnight.
+
+This tutorial is for developers who know the basics of Compact and have deployed a smart contract before. If you are new to Compact, read the [Compact language reference](/compact/reference/compact-reference) first or check out [Midnight Academy](https://academy.midnight.network).
+
+## How tokens work on Midnight
+
+Midnight has no native token live on the network yet. NIGHT exists on Cardano and bridges to Midnight later. Until that lands, a smart contract creates every token on Midnight — which is exactly what you do here.
```
oduameh: Em-dash. Consider replacing with a comma.

```suggestion
Midnight has no native token live on the network yet. NIGHT exists on Cardano and bridges to Midnight later. Until that lands, a smart contract creates every token on Midnight, which is exactly what you do here.
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+It helps to place your token in the wider map. Tokens fall into four types, along two axes: where the token comes from, and whether it is transparent or privacy-preserving.
+
+|  | Unshielded (transparent) | Shielded (privacy-preserving) |
+| :---- | :---- | :---- |
+| **Native** | Protocol-issued, for example NIGHT once bridged | Protocol-issued, amounts hidden |
+| **Contract** | Minted by a smart contract; balances on-chain | Minted by a smart contract; balances hidden |
+
+This tutorial covers the two **contract** rows. Native tokens are protocol-issued and are not yet live. A third group, hybrid tokens, places a native coin under the control of a smart contract; that work depends on protocol changes still in design, so it is out of scope here.
+
+:::info
```
oduameh: Admonitions should use bracket syntax for titles (style guide rule 55).

```suggestion
:::info[Token standards]
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+
+The token standards for Midnight are still taking shape. The smart contracts in this tutorial use the protocol primitives directly, which work today, rather than any draft standard. For the current direction, see the [token standards discussion](https://github.com/midnightntwrk/midnight-improvement-proposals/discussions/142).
+
+:::
+
+Those primitives share a starting point. Before you can mint, transfer, or check a balance, you need a way to identify the token you are working with — and that identifier is its color.
```
oduameh: Em-dash. Consider restructuring into two sentences.

```suggestion
Those primitives share a starting point. Before you can mint, transfer, or check a balance, you need a way to identify the token you are working with. That identifier is its color.
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+
+## Token color
+
+Every token has a *color*, a `Bytes<32>` value that identifies the token type. A contract token's color comes from the smart contract and a *domain separator*, a fixed label you choose. The minting primitive returns the color, and you use that color in every later operation on the token. Because the color depends on the smart contract, two smart contracts that use the same label still produce different tokens.
+
+Color settles which token you are working with, not who holds how much of it. How a contract keeps track of that — or whether it needs to at all — comes down to the privacy model.
```
oduameh: Em-dashes. Consider using parentheses for the aside.

```suggestion
Color settles which token you are working with, not who holds how much of it. How a contract keeps track of that (or whether it needs to at all) comes down to the privacy model.
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+
+On the **shielded** side, the chain hides balances. It records commitments and nullifiers, not amounts. There is no public balance to read. A smart contract that needs to know who holds what has to track that itself, and it has to guard the value as a shielded *coin*. That is why the shielded smart contract later mints a coin and delivers it to the caller in a single transaction, tracking no balances at all, while the unshielded one leans on the chain's public balances.
+
+|  | Unshielded token | Shielded token |
+| :---- | :---- | :---- |
+| Balance storage | On-chain, per address | Hidden; tracked by the smart contract |
+| Read a balance | `unshieldedBalance(color)` | No public read; smart contract state only |
+| Value the contract holds | A balance | An immutable coin |
+
+A shielded coin is immutable — you cannot change its value in place. To change the amount it holds, a smart contract consumes the current coin and creates a new one at the new value, and a nonce keeps each coin unique. The shielded tutorial's [Going further](/tokens/shielded-token#going-further-with-committed-transfers-and-burns) section shows this consume-and-recreate pattern.
```
oduameh: Em-dash. Consider splitting into two sentences.

```suggestion
A shielded coin is immutable. You cannot change its value in place. To change the amount it holds, a smart contract consumes the current coin and creates a new one at the new value. A nonce keeps each coin unique. The shielded tutorial's [Going further](/tokens/shielded-token#going-further-with-committed-transfers-and-burns) section shows this consume-and-recreate pattern.
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+
+Whether a token sits as a balance or a coin, at some point it moves from one holder to another. That movement takes one of two forms.
+
+## Two ways to transfer
+
+Tokens move in one of two ways, and the difference matters most for shielded tokens.
+
+1. **Wallet to wallet:** Once a holder has tokens in their wallet, the wallet transfers them directly. The smart contract is not involved.
+2. **Contract-mediated:** A smart contract moves tokens it holds — with `sendUnshielded` for a transparent token, or by sending a shielded coin for a shielded token.
```
oduameh: Em-dash. Consider replacing with a comma.

```suggestion
2. **Contract-mediated:** A smart contract moves tokens it holds, with `sendUnshielded` for a transparent token or by sending a shielded coin for a shielded token.
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+
+## Set up your project
+
+This section installs the toolchain, starts a local Midnight network, and lays out the project you build on across both parts of this tutorial. If your environment is already configured, skip ahead to [Build an unshielded token](#build-an-unshielded-token).
+
+### Prerequisites
+
+Before you begin, make sure you have:
+
+- The **Compact toolchain** installed — follow [Install the toolchain](/getting-started/installation). This tutorial targets compiler version 0.31.0.
```
oduameh: Em-dash. Consider splitting into separate sentences.

```suggestion
- The **Compact toolchain** installed. Follow [Install the toolchain](/getting-started/installation). This tutorial targets compiler version 0.31.0.
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+
+The network exposes:
+
+- Node: `http://localhost:9944`
+- Indexer: `http://localhost:8088`
+- Proof server: `http://localhost:6300`
+
+Because midnight-local-dev already runs a proof server, you do not start one separately.
+
+:::info Prefer Preprod?
```
oduameh: Admonition titles should use bracket syntax (style guide rule 55).

```suggestion
:::info[Prefer Preprod?]
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+docker run -p 6300:6300 midnightntwrk/proof-server:8.0.3 midnight-proof-server -v
+```
+
+and fund your own wallet at the Preprod faucet (covered in [Run it](#run-it)). The flow is identical; it just takes a couple of minutes more than local.
+
+:::
+
+### Create the project
+
+Create a project and install the Midnight dependencies. The versions below form a known-good set — install them exactly as written rather than letting npm resolve the latest.
```
oduameh: Em-dash. Consider splitting into two sentences.

```suggestion
Create a project and install the Midnight dependencies. The versions below form a known-good set. Install them exactly as written rather than letting npm resolve the latest.
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+npm install \
+  @midnight-ntwrk/midnight-js-contracts@4.1.1 \
+  @midnight-ntwrk/midnight-js-indexer-public-data-provider@4.1.1 \
+  @midnight-ntwrk/midnight-js-http-client-proof-provider@4.1.1 \
+  @midnight-ntwrk/midnight-js-node-zk-config-provider@4.1.1 \
+  @midnight-ntwrk/midnight-js-types@4.1.1 \
+  @midnight-ntwrk/midnight-js-utils@4.1.1
+```
+
+You add the wallet packages in the deploy section, where you connect to the network.
```
oduameh: Em-dash. Consider splitting into a new sentence.

```suggestion
The steps below use the **local** network by default. To run against **Preprod** instead, set `MIDNIGHT_NETWORK=preprod` when you run the script. Nothing else changes.
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+
+Start the file with the language version and the standard library:
+
+```compact
+pragma language_version 0.23;
+
+import CompactStandardLibrary;
+```
+
+The pragma pins the Compact language version, so the compiler rejects the file if it does not match.
```
oduameh: Em-dash. Consider splitting into two sentences.

```suggestion
The `wallet-sdk` version matters here. Pin it to 1.0.0. A newer minor pulls in a ledger version that the network rejects when you deploy a contract.
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+### Declare the on-chain state
+
+Add the contract's public state:
+
+```compact
+export ledger token_color: Bytes<32>;
+export ledger initialized: Boolean;
+```
+
+A ledger field lives on-chain. `token_color` holds the token's identity, which you set the first time you mint; `initialized` records whether that has happened yet, so the other circuits can check it before acting. On-chain ledger state is public and readable by anyone either way — what `export` adds is the generated JavaScript binding in the compiled artifacts, so your deploy script can read these fields back by name.
```
oduameh: Em-dash. Consider splitting into a new sentence.

```suggestion
A ledger field lives on-chain. `token_color` holds the token's identity, which you set the first time you mint; `initialized` records whether that has happened yet, so the other circuits can check it before acting. On-chain ledger state is public and readable by anyone either way. What `export` adds is the generated JavaScript binding in the compiled artifacts, so your deploy script can read these fields back by name.
```

oduameh (commented):
Nice work on this PR. These are substantial, well-structured tutorials that walk developers through both unshielded and shielded tokens end to end. The progression from transparent balances to immutable coins is well motivated and the code examples are thorough.

Left inline suggestions that can be batch committed.

Top recurring issues:
- **Em-dashes** appear throughout both tutorials. The Midnight docs style guide uses commas, periods, colons, or restructured sentences instead. Each instance below has an inline fix.
- **Compiler version**: both tutorials reference compactc 0.31.0. Worth verifying against the current compatibility matrix before merging.
- **Intro scope**: the unshielded tutorial intro says it covers both tokens, but only this file covers the unshielded one.

Additional notes (not inlined):
- The shielded tutorial's deploy script duplicates the wallet-funding flow from the unshielded tutorial nearly verbatim (~160 lines). Consider extracting the shared setup into a partial or collapsible block to reduce maintenance surface.
- "Midnight has no native token live on the network yet" (unshielded-token.mdx, line 21): with Mainnet launched, this may need qualifying (e.g. 'native mNIGHT is not yet live').
- The genesis wallet seed `0x00...001` is shown in the run command. Consider adding a note that this seed is for local use only.

27 suggestions, 1 comment across 2 files.

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+  token_color = color;
+  initialized = true;
+}
+```
+
+In the code snippet above:
+
+- `pad(32, "tutorial:unshielded:token")` builds the *domain separator* with a fixed `Bytes<32>` label that, with the contract's address, determines the token color.
+- `mintUnshieldedToken(...)` mints `amount` units and returns the token's color. The recipient is `kernel.self()`, the contract's own address, wrapped in `left<ContractAddress, UserAddress>(...)` to mark it as the contract side of the address type. So the new supply lands with the contract.
+- The last two lines record the color and set `initialized` to true. Two things are worth noting. The mint amount is a `Uint<64>` — the protocol caps a single mint at that width, while transfers use the wider `Uint<128>`. And `disclose(...)` marks a value as safe to reveal on-chain, and minting is a public action, so you disclose the amount.
```
oduameh: Em-dash. Consider using parentheses.

```suggestion
- The last two lines record the color and set `initialized` to true. Two things are worth noting. The mint amount is a `Uint<64>` (the protocol caps a single mint at that width), while transfers use the wider `Uint<128>`. And `disclose(...)` marks a value as safe to reveal on-chain, and minting is a public action, so you disclose the amount.
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+
+Minting to `kernel.self()` credits the contract directly, so the new supply is held by the contract until you transfer it out.
+
+This mint is permissionless — anyone can call it. That suits a faucet-style test token with no real value. A token that carries value needs an authorization check here, which you can add as an extension. The [shielded tutorial](/tokens/shielded-token) shows one way to do this.
```
oduameh: Em-dash. Consider using a colon.

```suggestion
This mint is permissionless: anyone can call it. That suits a faucet-style test token with no real value. A token that carries value needs an authorization check here, which you can add as an extension. The [shielded tutorial](/tokens/shielded-token) shows one way to do this.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+
+A freshly minted coin has no Merkle tree position yet, so it cannot be spent with `sendShielded`. Once the ledger commits it on-chain, it gains an `mtIndex` and becomes a `QualifiedShieldedCoinInfo`, which is what `sendShielded` requires. This tutorial mints and spends a coin in the *same* transaction, which keeps you on the fresh-coin path and avoids tracking Merkle positions across transactions. [Going further](#going-further-with-committed-transfers-and-burns) covers the committed path.
+
+One more constraint shapes what a contract-mediated shielded transfer can do today. The `sendShielded` primitive does not currently create coin ciphertexts, so sending shielded value to a user other than the current caller will not notify that user of the coin. In practice that means a contract can reliably deliver shielded value to the **caller** (the person signing the transaction). That is the transfer this tutorial demonstrates.
+
+## Set up
+
+You reuse the unshielded tutorial's environment. In particular:
+
+- The toolchain (compactc 0.31.0), the local network (midnight-local-dev), and the proof server — see the [project setup](/tokens/unshielded-token#set-up-your-project).
```
oduameh: Em-dash. Consider splitting into two sentences. (Also worth verifying compactc version against the current compatibility matrix.)

```suggestion
- The toolchain (compactc 0.31.0), the local network (midnight-local-dev), and the proof server. See the [project setup](/tokens/unshielded-token#set-up-your-project).
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+- `src/config.ts` and `src/wallet.ts` — **identical to the unshielded tutorial**, copy them over unchanged.
```
oduameh: Em-dash. Consider restructuring with a comma and new sentence.

```suggestion
- `src/config.ts` and `src/wallet.ts`, **identical to the unshielded tutorial**. Copy them over unchanged.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+```compact
+pragma language_version 0.23;
+
+import CompactStandardLibrary;
+
+witness localNonceSeed(): Bytes<32>;
+witness minterSecretKey(): Bytes<32>;
+```
+
+The new line is the witnesses. A witness is a function that runs off-chain and feeds a value into the circuit. That value is often private data the circuit uses without publishing it, but it doesn't have to be — a witness is really just a hook for off-chain computation, and what it returns can be public or private. Privacy isn't tied to witnesses either: a value passed as an ordinary circuit parameter also stays private unless the circuit explicitly discloses it.
```
oduameh: Em-dash, and the sentence is 38 words. Consider splitting into shorter sentences.

```suggestion
The new line is the witnesses. A witness is a function that runs off-chain and feeds a value into the circuit. That value is often private data the circuit uses without publishing it, but it does not have to be. A witness is a hook for off-chain computation, and what it returns can be public or private. Privacy is not tied to witnesses either: a value passed as an ordinary circuit parameter also stays private unless the circuit explicitly discloses it.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+### Declare the on-chain state
+
+```compact
+export ledger token_color: Bytes<32>;
+export ledger initialized: Boolean;
+export ledger mints: Counter;
+export ledger minter: Bytes<32>;
+```
+
+`token_color` and `initialized` play the same role as in the unshielded tutorial. Note what is *not* here: there is no balance, and no map of who-holds-what. Recording shielded balances on-chain would publish the very amounts shielding is meant to hide, so the contract keeps none. `mints` is a simple counter, used only to show on-chain state changing.
```
oduameh: "simple" is flagged as marketing language (style guide rule 12). Consider dropping it.

```suggestion
`token_color` and `initialized` play the same role as in the unshielded tutorial. Note what is *not* here: there is no balance, and no map of who-holds-what. Recording shielded balances on-chain would publish the very amounts shielding is meant to hide, so the contract keeps none. `mints` is a counter, used only to show on-chain state changing.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+}
+
+// Derive a public key from the secret with a domain-separated hash. A caller can
+// reproduce this value only if they hold the secret, so it is safe to gate on.
+export circuit deriveMinterKey(sk: Bytes<32>): Bytes<32> {
+  return persistentHash<Vector<2, Bytes<32>>>([pad(32, "tutorial:shielded:minter:v1"), sk]);
+}
+```
+
+`deriveMinterKey` hashes a secret with a domain separator into a public key. The constructor runs once at deploy and records the *deployer's* derived key as the authorized minter, so only the wallet holding the deploy secret can mint later. (`ownPublicKey()` is deliberately not used for this — it returns a prover-claimed value with no binding to the signer, so any check on it is bypassable.) As in the unshielded tutorial, the constructor doesn't set `token_color` — that's set when you mint — and `initialized` starts false.
```
oduameh: Multiple em-dashes on this line. Consider restructuring with periods and parentheses.

```suggestion
`deriveMinterKey` hashes a secret with a domain separator into a public key. The constructor runs once at deploy and records the *deployer's* derived key as the authorized minter, so only the wallet holding the deploy secret can mint later. (`ownPublicKey()` is deliberately not used for this. It returns a prover-claimed value with no binding to the signer, so any check on it is bypassable.) As in the unshielded tutorial, the constructor does not set `token_color` (that is set when you mint), and `initialized` starts false.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+- `pad(32, "tutorial:shielded:token")` builds the domain separator, exactly as in the unshielded tutorial, but with a label of its own so the shielded token gets a distinct color.
+- `evolveNonce(nonceIndex, localNonceSeed())` derives a unique nonce. The seed is private (from the witness); `disclose(...)` wraps the *derived* nonce because the mint publishes it, while the seed stays secret. You pass a different `nonceIndex` each time you mint so the nonces never repeat.
+- `mintShieldedToken(...)` mints `amount` units and returns a `ShieldedCoinInfo`, a fresh coin. The recipient `right<ZswapCoinPublicKey, ContractAddress>(kernel.self())` is the contract's own address, marked as the contract side of the recipient type, so the new coin belongs to the contract.
+- `sendImmediateShielded(coin, ..., coin.value)` spends that fresh coin in the same transaction, sending the whole value to `ownPublicKey()` — the caller. You wrap the recipient with `left<ZswapCoinPublicKey, ContractAddress>(...)` because it is a user key, not a contract. Because you send the full value, there is no change coin to manage.
```
oduameh: Em-dash. Consider replacing with a comma.

```suggestion
- `sendImmediateShielded(coin, ..., coin.value)` spends that fresh coin in the same transaction, sending the whole value to `ownPublicKey()`, the caller. You wrap the recipient with `left<ZswapCoinPublicKey, ContractAddress>(...)` because it is a user key, not a contract. Because you send the full value, there is no change coin to manage.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+
+That completes the contract. Compile it the same way as before, into its own output directory:
+
+```bash
+compact compile contracts/shielded-token.compact src/shielded/managed/shielded-token
+```
+
+## Wire the contract for deployment
+
+All of the shielded TypeScript goes in `src/shielded/`. Two of the support files differ from the unshielded tutorial; you reuse the other two — `config.ts` and `wallet.ts` — from `src/` unchanged, which is why the imports below reach up a level with `../`.
```
oduameh: Em-dashes. Consider replacing with commas or restructuring.

```suggestion
All of the shielded TypeScript goes in `src/shielded/`. Two of the support files differ from the unshielded tutorial. You reuse `config.ts` and `wallet.ts` from `src/` unchanged, which is why the imports below reach up a level with `../`.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+      config.proofServer,
+      zkConfigProvider,
+    ),
+    walletProvider: wallet,
+    midnightProvider: wallet,
+  };
+}
+```
+
+Unlike the unshielded tutorial, the private state is not empty — it carries the nonce seed. You pass the real value at deploy time in the next step.
```
oduameh: Em-dash. Consider splitting into two sentences.

```suggestion
Unlike the unshielded tutorial, the private state is not empty. It carries the nonce seed. You pass the real value at deploy time in the next step.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+
+## Deploy and test
+
+The deploy script follows the same shape as before — build the wallet, wait for funds and DUST, then deploy — and differs only in what it deploys and calls. The wallet, funding, and DUST steps match the unshielded tutorial, but this tutorial includes them in full so it runs on its own.
```
oduameh: Em-dashes. Consider using parentheses.

```suggestion
The deploy script follows the same shape as before (build the wallet, wait for funds and DUST, then deploy) and differs only in what it deploys and calls. The wallet, funding, and DUST steps match the unshielded tutorial, but this tutorial includes them in full so it runs on its own.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+});
+
+const afterMint = await readLedger();
+logger.info(`initialized after mint: ${afterMint.initialized}`); // true
+logger.info(`token color: ${toHex(afterMint.token_color)}`);
+```
+
+`mint_and_send` mints 1,000 shielded units and sends them to the caller in the same transaction. `nonceIndex` is `0n` for this first call; a second call would use `1n`, and so on, to keep nonces unique.
+
+Finally, confirm the transfer by reading your **wallet's** shielded balance — there is no contract balance to read for a shielded token:
```
oduameh: Em-dash. Consider splitting into two sentences.

```suggestion
Finally, confirm the transfer by reading your **wallet's** shielded balance. There is no contract balance to read for a shielded token:
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+```typescript
+const after = await syncWallet(logger, wallet.wallet, 60 * 60_000);
+const color = toHex(afterMint.token_color);
+const shieldedBalance = after.shielded.balances[color] ?? 0n;
+logger.info(`wallet shielded balance of the token: ${shieldedBalance}`); // expect 1000
+
+await wallet.stop();
+```
+
+After syncing, the wallet should report 1,000 of the token's color in its shielded balance — proof that the contract minted shielded value and delivered it to you privately.
```
oduameh: Em-dash. Consider splitting into two sentences.

```suggestion
After syncing, the wallet should report 1,000 of the token's color in its shielded balance. This confirms the contract minted shielded value and delivered it to you privately.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+
+```bash
+MIDNIGHT_NETWORK=local \
+MIDNIGHT_SEED=0000000000000000000000000000000000000000000000000000000000000001 \
+npx tsx src/shielded/deploy.ts
+```
+
+You should see the contract address, `initialized` flipping to true after the mint, and a wallet shielded balance of 1,000. To run against Preprod instead, use the same Preprod instructions as the [unshielded tutorial](/tokens/unshielded-token#run-against-preprod-instead).
+
+## Going further with committed transfers and burns
```
oduameh: Prefer imperative verbs over gerunds in task-oriented headings (style guide rule 32).

```suggestion
## Go further with committed transfers and burns
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+## What you built
+
+You created a shielded token and moved it, end to end:
+
+- `mintShieldedToken` creates a fresh shielded coin owned by the contract.
+- `evolveNonce`, backed by a private witness seed, derives unique coin nonces.
+- `sendImmediateShielded` spends that fresh coin in the same transaction and delivers it to the caller.
+- Because shielded balances are not public, you confirmed the result by reading the wallet's shielded balance rather than the contract's.
+
+The thread connecting the two tutorials is that privacy is not a setting — it changes what the contract stores and how value moves. The [unshielded token](/tokens/unshielded-token) leaned on the chain to track balances; the shielded token holds value in coins and never publishes amounts at all.
```
oduameh: Em-dash. Consider splitting into two sentences.

```suggestion
The thread connecting the two tutorials is that privacy is not a setting. It changes what the contract stores and how value moves. The [unshielded token](/tokens/unshielded-token) leaned on the chain to track balances; the shielded token holds value in coins and never publishes amounts at all.
```

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+title: Create and transfer an unshielded token
+description: "Build, deploy, and transfer a transparent (unshielded) token in Compact on Midnight, with on-chain balances you read straight from wallet state."
+sidebar_label: Unshielded token
+sidebar_position: 2
+tags: [tokens, unshielded, compact, tutorial, mint, transfer]
+---
+
+# Create and transfer an unshielded token
+
+On Midnight, the privacy model you choose for a token changes how you write its smart contract. This tutorial shows you how to create and transfer both **unshielded** (transparent) and **shielded** (privacy-preserving) tokens in Compact.
```
oduameh: This intro says the tutorial covers "both unshielded and shielded tokens," but this file only covers the unshielded token. Would it be worth scoping the intro to this file and pointing readers to the companion tutorial?

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+**There is no public balance.** The chain records commitments and nullifiers, not amounts. A contract cannot call anything like `unshieldedBalance(color)` for shielded value, because that number does not exist on-chain. To know what it holds, a contract has to track its own coins; to prove a transfer worked, you read the *wallet's* shielded balance, not the contract's.
+
+**Value lives in immutable coins.** A `ShieldedCoinInfo` describes a shielded coin: a nonce, a color, and a value. You cannot change a coin's value in place. To move part of it, you spend the coin and the standard library hands you back a new *change* coin for the remainder. A nonce keeps every coin unique, and you derive nonces deterministically with `evolveNonce` from a private seed.
+
+Fresh coins and committed coins are spent differently:
+
+| Coin state | Type | How you spend it |
+| :---- | :---- | :---- |
+| Fresh (just created, same transaction) | `ShieldedCoinInfo` | `sendImmediateShielded` |
+| Committed (already in the ledger) | `QualifiedShieldedCoinInfo` (adds `mtIndex`) | `sendShielded` |
```
oduameh: The actual source defines this field as `mt_index` (snake_case). Verified against `QualifiedShieldedCoinInfo` in the standard library.

```suggestion
| Committed (already in the ledger) | `QualifiedShieldedCoinInfo` (adds `mt_index`) | `sendShielded` |
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+
+A freshly minted coin has no Merkle tree position yet, so it cannot be spent with `sendShielded`. Once the ledger commits it on-chain, it gains an `mtIndex` and becomes a `QualifiedShieldedCoinInfo`, which is what `sendShielded` requires. This tutorial mints and spends a coin in the *same* transaction, which keeps you on the fresh-coin path and avoids tracking Merkle positions across transactions. [Going further](#going-further-with-committed-transfers-and-burns) covers the committed path.
```
oduameh: Same field name discrepancy: source uses `mt_index`, not `mtIndex`.

```suggestion
A freshly minted coin has no Merkle tree position yet, so it cannot be spent with `sendShielded`. Once the ledger commits it on-chain, it gains an `mt_index` and becomes a `QualifiedShieldedCoinInfo`, which is what `sendShielded` requires. This tutorial mints and spends a coin in the *same* transaction, which keeps you on the fresh-coin path and avoids tracking Merkle positions across transactions. [Going further](#going-further-with-committed-transfers-and-burns) covers the committed path.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+npx tsx src/shielded/deploy.ts
+```
+
+You should see the contract address, `initialized` flipping to true after the mint, and a wallet shielded balance of 1,000. To run against Preprod instead, use the same Preprod instructions as the [unshielded tutorial](/tokens/unshielded-token#run-against-preprod-instead).
+
+## Going further with committed transfers and burns
+
+The flow above mints and spends a coin in one transaction, which keeps you on the fresh-coin path. Two common operations need the committed path or a special recipient. Their primitives are part of the same standard library; the signatures below are exact, but wiring them end-to-end on a live node involves managing Merkle positions across transactions, so treat them as the next thing to build rather than copy-paste-ready.
+
+1. **Transferring a coin the contract already holds.** Once the ledger commits a minted coin on-chain, it becomes a `QualifiedShieldedCoinInfo` (it gains a Merkle tree index, `mtIndex`), and you spend it with `sendShielded`:
```
oduameh: Same `mt_index` fix.

```suggestion
1. **Transferring a coin the contract already holds.** Once the ledger commits a minted coin on-chain, it becomes a `QualifiedShieldedCoinInfo` (it gains a Merkle tree index, `mt_index`), and you spend it with `sendShielded`:
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+circuit sendShielded(
+  input: QualifiedShieldedCoinInfo,
+  recipient: Either<ZswapCoinPublicKey, ContractAddress>,
+  value: Uint<128>
+): ShieldedSendResult;
+```
+
+`sendShielded` returns a `ShieldedSendResult { change: Maybe<ShieldedCoinInfo>; sent: ShieldedCoinInfo; }`. If you send less than the coin's value, `change` holds a new coin for the remainder — this is the consume-and-recreate pattern, and your application must persist that change coin, wait for it to commit, and spend it later as its own `QualifiedShieldedCoinInfo`.
+
+The hard part on a live network is obtaining the `mtIndex` for the committed coin and supplying it to the circuit, typically from the indexer; that is the piece to work out and test carefully. Remember too that delivering shielded value to a user other than the caller will not notify them today.
```
oduameh: Same `mt_index` fix.

```suggestion
The hard part on a live network is obtaining the `mt_index` for the committed coin and supplying it to the circuit, typically from the indexer; that is the piece to work out and test carefully. Remember too that delivering shielded value to a user other than the caller will not notify them today.
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+| :---- | :---- | :---- |
+| Fresh (just created, same transaction) | `ShieldedCoinInfo` | `sendImmediateShielded` |
+| Committed (already in the ledger) | `QualifiedShieldedCoinInfo` (adds `mtIndex`) | `sendShielded` |
+
+A freshly minted coin has no Merkle tree position yet, so it cannot be spent with `sendShielded`. Once the ledger commits it on-chain, it gains an `mtIndex` and becomes a `QualifiedShieldedCoinInfo`, which is what `sendShielded` requires. This tutorial mints and spends a coin in the *same* transaction, which keeps you on the fresh-coin path and avoids tracking Merkle positions across transactions. [Going further](#going-further-with-committed-transfers-and-burns) covers the committed path.
+
+One more constraint shapes what a contract-mediated shielded transfer can do today. The `sendShielded` primitive does not currently create coin ciphertexts, so sending shielded value to a user other than the current caller will not notify that user of the coin. In practice that means a contract can reliably deliver shielded value to the **caller** (the person signing the transaction). That is the transfer this tutorial demonstrates.
```
oduameh, line 36: This is a protocol-level behavioral claim (coin ciphertexts, recipient notification) that cannot be verified from the standard library source alone. Recommend having a protocol engineer confirm before merging.

docs/tokens/shielded-token.mdx
oduameh: `ownPublicKey()` is a compiler built-in, not defined in the standard library. The claim that it returns a prover-claimed value with no binding to the signer is architecturally plausible (ZK circuits cannot enforce who the prover is), but there is no source-level evidence in the standard library to confirm or refute this exact framing. This is the highest-risk claim in the PR. Recommend protocol engineer review.

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+
+```compact
+circuit sendShielded(
+  input: QualifiedShieldedCoinInfo,
+  recipient: Either<ZswapCoinPublicKey, ContractAddress>,
+  value: Uint<128>
+): ShieldedSendResult;
+```
+
+`sendShielded` returns a `ShieldedSendResult { change: Maybe<ShieldedCoinInfo>; sent: ShieldedCoinInfo; }`. If you send less than the coin's value, `change` holds a new coin for the remainder — this is the consume-and-recreate pattern, and your application must persist that change coin, wait for it to commit, and spend it later as its own `QualifiedShieldedCoinInfo`.
```
oduameh: `sendShielded`'s return type is documented here, but `sendImmediateShielded` also returns `ShieldedSendResult`, not `[]`. The tutorial ignores the return value (valid Compact), but a developer doing a partial send (less than `coin.value`) needs the change coin from the result. Worth noting that both primitives share the same return type.

docs/tokens/_category_.yaml
```diff
@@ -3,7 +3,8 @@ label: 'Tokens'
 collapsible: true
 collapsed: true
 link:
-  type: doc
-  id: index
+  type: generated-index
```
oduameh, line 6: Switching from `type: doc` / `id: index` to `generated-index` changes the URL behavior. `/tokens` now shows an auto-generated index page rather than the hand-authored overview (which moves to `/tokens/overview`). Verify this is the intended behavior and that any existing links to `/tokens` still land where expected.

docs/tokens/_category_.yaml
```diff
@@ -3,7 +3,8 @@ label: 'Tokens'
...
+  title: Tokens
+  description: NIGHT, DUST, shielded and unshielded tokens, custom tokens, fees, and how to create and transfer your own tokens on Midnight.
```
oduameh, line 9: This `description` differs from the `customProps.description` on line 10 (this one adds "and how to create and transfer your own tokens on Midnight"). Consider syncing them or removing one to avoid divergence over time.

docs/tokens/unshielded-token.mdx
oduameh: compactc 0.31.0 is correct today, but 0.31.1 is now available. The language version `0.23` is correct. Readers following this tutorial later may have a different patch version; worth noting or linking to the compatibility matrix.

oduameh (commented):
Source-verified technical review. Every Compact standard library claim was checked against the actual source at `LFDT-Minokawa/compact/compiler/standard-library.compact`.

All function signatures confirmed (mintUnshieldedToken, sendUnshielded, unshieldedBalanceGte, mintShieldedToken, sendShielded, sendImmediateShielded, evolveNonce, shieldedBurnAddress, pad, persistentHash, all Either type wrapping). One field name discrepancy and two protocol-level claims that cannot be verified from the standard library source are flagged below.

Structural notes:
- The shielded tutorial's deploy script duplicates ~160 lines of wallet-funding flow from the unshielded tutorial. The PR notes this is intentional ("included here in full so it runs on its own"), but it creates a maintenance burden. Consider a collapsible block or partial.
- Renaming `index.mdx` to `overview.mdx` and switching to `generated-index` changes the URL from `/tokens` to `/tokens/overview`. Any existing links to `/tokens` now land on the auto-generated index page. Flagged inline on `_category_.yaml`.

4 suggestions, 7 comments across 3 files.

docs/tokens/_category_.yaml
Olanetsoft: Yes, thats intended, i have fixed `/tokens` redirect.

Olanetsoft:
> Source-verified technical review. Every Compact standard library claim was checked against the actual source at `LFDT-Minokawa/compact/compiler/standard-library.compact`.
> 
> All function signatures confirmed (mintUnshieldedToken, sendUnshielded, unshieldedBalanceGte, mintShieldedToken, sendShielded, sendImmediateShielded, evolveNonce, shieldedBurnAddress, pad, persistentHash, all Either type wrapping). One field name discrepancy and two protocol-level claims that cannot be verified from the standard library source are flagged below.
> 
> Structural notes:
> 
> * The shielded tutorial's deploy script duplicates ~160 lines of wallet-funding flow from the unshielded tutorial. The PR notes this is intentional ("included here in full so it runs on its own"), but it creates a maintenance burden. Consider a collapsible block or partial.
> * Renaming `index.mdx` to `overview.mdx` and switching to `generated-index` changes the URL from `/tokens` to `/tokens/overview`. Any existing links to `/tokens` now land on the auto-generated index page. Flagged inline on `_category_.yaml`.
> 
> 4 suggestions, 7 comments across 3 files.

All fixed. For the snippet that seem to be verbatim, its intentional. Managing should be easy because update to tokens should cover both shielded and unshielded.

docs/tokens/shielded-token.mdx
Olanetsoft: Verified against source, this is confirmed, no protocol-engineer sign-off needed.

The upstream Compact stdlib doc states it directly ([CompactStandardLibrary/exports.md](https://github.com/LFDT-Minokawa/compact/blob/main/doc/api/CompactStandardLibrary/exports.md#L687)):

`this does not currently create coin ciphertexts, so sending to a user public key except for the current user will not lead to this user being informed of the coin they've been sent.`

Corroborated end-to-end in source:

- sendShielded only calls createZswapOutput(output, recipient), no encryption ([standard-library.compact:161](https://github.com/LFDT-Minokawa/compact/blob/main/compiler/standard-library.compact#L161)); the runtime output type is { coinInfo, recipient }, no ciphertext field ([zswap.ts:255](https://github.com/LFDT-Minokawa/compact/blob/main/runtime/src/zswap.ts#L255)).
- Ledger Output.ciphertext is created only when a recipient encryption key is supplied ([construct.rs:232](https://github.com/midnightntwrk/midnight-ledger/blob/main/zswap/src/construct.rs#L232)); the contract-owned path passes None, and the on-chain Effects struct has no field to supply one ([context.rs:645](https://github.com/midnightntwrk/midnight-ledger/blob/main/onchain-runtime/src/context.rs#L645)).
- Wallet scanning skips None-ciphertext outputs ([zswap_keys.rs:289](https://github.com/midnightntwrk/midnight-ledger/blob/main/ledger-wasm/src/zswap_keys.rs#L289)), so a non-caller recipient is never notified while the caller holds the coin preimage locally and needs no ciphertext. Resolving.

oduameh (dismissed):

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+};
+
+export const PREVIEW_CONFIG: NetworkConfig = {
+  networkId: 'preview',
+  indexer: 'https://indexer.preview.midnight.network/api/v4/graphql',
+  indexerWS: 'wss://indexer.preview.midnight.network/api/v4/graphql/ws',
+  node: 'https://rpc.preview.midnight.network',
+  nodeWS: 'wss://rpc.preview.midnight.network',
+  proofServer: process.env['MIDNIGHT_PROOF_SERVER'] ?? 'http://127.0.0.1:6300',
+  faucet: 'https://faucet.preview.midnight.network/api/drips',
```
nstanford5: These are the faucet links that are being deprecated -- please update to nethermind faucets

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+export circuit transfer(recipient: UserAddress, amount: Uint<128>): [] {
+  assert(initialized, "token not minted yet");
+  assert(unshieldedBalanceGte(token_color, disclose(amount)), "insufficient balance");
+  sendUnshielded(
+    token_color,
+    disclose(amount),
+    right<ContractAddress, UserAddress>(disclose(recipient))
+  );
+}
+```
```
nstanford5, line 247: This circuit either needs auth or needs a warning that anyone can call this and drain it to any address they like.

While this tutorial is about Tokens specifically, it would be too easy for a human (or an AI) to copy and paste this code assuming that it is safe. My preference would be to add auth as if this were a real token, rather than having this all be permissionless

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+export const CompiledUnshieldedToken = CompiledContract.make(
+  'UnshieldedToken',
+  Contract,
+).pipe(
+  CompiledContract.withVacantWitnesses,
+  CompiledContract.withCompiledFileAssets(zkConfigPath),
+);
+```
+
+`withVacantWitnesses` applies because the contract has no witnesses; its state is entirely public. `zkConfigPath` points the proof server at the keys the compiler generated under `src/managed/unshielded-token`.
```
nstanford5: Witness !== private state

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+    wallet.unshieldedKeystore.getPublicKey(),
+    (payload: Uint8Array) => wallet.unshieldedKeystore.signData(payload),
+  );
+  const finalized = await wallet.wallet.finalizeRecipe(recipe);
+  const txId = await wallet.wallet.submitTransaction(finalized);
+  logger.info(`DUST registration submitted: ${txId}`);
+} else {
+  logger.info('NIGHT is already registered for DUST generation.');
+}
+```
```
nstanford5, line 781: Is this checking for a NIGHT UTxO specifically? This crashes on a re-run with a wallet that already holds the test token. Likely because this is taking any UTxO instead of filtering for NIGHT first

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+
+On Midnight, the privacy model you choose for a token changes how you write its smart contract. This tutorial shows you how to create and transfer an **unshielded** (transparent) token in Compact: one whose balances live on-chain, where anyone can read them.
+
+This is the first of a two-part pair. Once you have built the transparent token here, the companion tutorial, [Create and transfer a shielded token](/tokens/shielded-token), builds the privacy-preserving version, where balances stay hidden and the design changes to match. Reading them in order is the fastest way to understand how privacy refines a token on Midnight.
+
+This tutorial is for developers who know the basics of Compact and have deployed a smart contract before. If you are new to Compact, read the [Compact language reference](/compact/reference/compact-reference) first or check out [Midnight Academy](https://academy.midnight.network).
+
+## How tokens work on Midnight
+
+Midnight has no native token live on the network yet. NIGHT exists on Cardano and bridges to Midnight later. Until that lands, a smart contract creates every token on Midnight, which is exactly what you do here.
```
nstanford5: This is only true for mainnet. The tests below demonstrate working with NIGHT on other networks.

docs/tokens/unshielded-token.mdx
```diff
@@ -0,0 +1,908 @@
...
+  if (dustBalance > 0n) break;
+  await new Promise((r) => setTimeout(r, 15_000));
+}
+if (dustBalance <= 0n) {
+  throw new Error('Timed out waiting for DUST to be generated.');
+}
+logger.info(`DUST available: ${dustBalance}`);
+```
+
+DUST generates from registered NIGHT over a short period, so the balance starts at 0 and climbs. The script polls every 15 seconds until it goes positive. On local this takes a few seconds.
```
nstanford5: How long does it take on preprod?

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+
+A freshly minted coin has no Merkle tree position yet, so it cannot be spent with `sendShielded`. Once the ledger commits it on-chain, it gains an `mt_index` and becomes a `QualifiedShieldedCoinInfo`, which is what `sendShielded` requires. This tutorial mints and spends a coin in the *same* transaction, which keeps you on the fresh-coin path and avoids tracking Merkle positions across transactions. [Go further](#go-further-with-committed-transfers-and-burns) covers the committed path.
+
+One more constraint shapes what a contract-mediated shielded transfer can do today. The `sendShielded` primitive does not currently create coin ciphertexts, so sending shielded value to a user other than the current caller will not notify that user of the coin. In practice that means a contract can reliably deliver shielded value to the **caller** (the person signing the transaction). That is the transfer this tutorial demonstrates.
+
+## Set up
+
+You reuse the unshielded tutorial's environment. In particular:
+
+- The toolchain (compactc 0.31.0), the local network (midnight-local-dev), and the proof server. See the [project setup](/tokens/unshielded-token#set-up-your-project).
```
nstanford5: We don't recommend this version due to the security flaw -- should be 0.31.1

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+  readonly nonceSeed: Uint8Array;
+  readonly minterSecret: Uint8Array;
+};
+
+export const createShieldedTokenPrivateState = (
+  nonceSeed: Uint8Array = createHash('sha256')
+    .update('tutorial:shielded:token:seed')
+    .digest(),
+  minterSecret: Uint8Array = createHash('sha256')
+    .update('tutorial:shielded:minter:secret')
```
nstanford5: With the high likelihood of a user (or AI) copy / pasting this code, we should avoid printing secrets or seeds like this. Other examples call `randomBytes(32)`

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+      config.proofServer,
+      zkConfigProvider,
+    ),
+    walletProvider: wallet,
+    midnightProvider: wallet,
+  };
+}
+```
+
+Unlike the unshielded tutorial, the private state is not empty. It carries the nonce seed. You pass the real value at deploy time in the next step.
```
nstanford5: It carries the nonce seed and minter secret

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+}
+if (dustBalance <= 0n) {
+  throw new Error('Timed out waiting for DUST to be generated.');
+}
+logger.info(`DUST available: ${dustBalance}`);
+```
+
+On the local network the genesis wallet is already funded and registered, so the NIGHT wait returns immediately, registration reports it is already done, and DUST appears within a few seconds.
+
+Now the shielded-specific part. Build the providers, deploy the contract with the nonce-seed private state, and read the initial state:
```
nstanford5:
```suggestion
Now the shielded-specific part. Build the providers, deploy the contract with the private state, and read the initial state:
```

docs/tokens/shielded-token.mdx
```diff
@@ -0,0 +1,534 @@
...
+
+async function readLedger() {
+  const state = await providers.publicDataProvider.queryContractState(contractAddress);
+  return ledger(state!.data);
+}
+
+logger.info(`initialized at deploy: ${(await readLedger()).initialized}`); // false
+```
+
+The one difference from the unshielded tutorial is `initialPrivateState` instead of `{}`: you pass `createShieldedTokenPrivateState()`, which seeds the witnesses.
```
nstanford5: It doesn't seed the witnesses, it populates the private state. Witness !== private state. In this case, the witnesses are just off-chain functions being used to access the private state. 

This is not entirely necessary, but offers one example with witnesses and one without. We need to be more clear about the purpose and power of witness functions

nstanford5 (changes requested):

docs/tokens/unshielded-token.mdx
Olanetsoft: Updated to filter.

docs/tokens/unshielded-token.mdx
Olanetsoft: Added

nstanford5 (approved):

PR #1126, opened 2026-07-10
docs: add browser dApp guide for levelPrivateStateProvider with canary password verification

pplmaverick:
## Summary

Adds a browser usage guide for `levelPrivateStateProvider`, covering:

1. How `cryptoBackend` auto-detection works in browser environments
2. Why a canary pattern is needed for early password validation
3. How to implement `resetStorage()` for production UX

## Motivation

Browser compatibility was added in midnight-js #827, but no documentation or example
exists showing how to use `levelPrivateStateProvider` in a browser dApp.

Additionally, the SDK's internal `verifyPassword()` only performs an in-memory cache
consistency check — it does not validate the password against stored encrypted data.
On first unlock (or after page reload), there is no password verification at all.
A wrong password causes an AES-GCM auth tag failure deep inside a later operation,
producing an opaque WebCrypto `OperationError` instead of a clear message at unlock time.

The canary pattern below provides fail-fast password validation at unlock time.

### Using levelPrivateStateProvider in browser dApps

```typescript
import { levelPrivateStateProvider }
  from '@midnight-ntwrk/midnight-js-level-private-state-provider';

// cryptoBackend auto-detects Web Crypto availability.
// In browsers, 'webcrypto' is selected automatically via isWebCryptoAvailable().
// Set it explicitly if you need predictable behavior across environments.
const provider = levelPrivateStateProvider({
  accountId: address,
  cryptoBackend: 'webcrypto', // explicit, recommended for browser dApps
});
```

### Canary pattern for early password validation

Without a canary, a wrong password isn't caught until the app attempts to decrypt
real private state data. AES-GCM auth tag verification then throws an opaque
`OperationError` deep inside a transaction flow rather than at unlock time.

A canary value written on first unlock lets you catch this at the earliest possible
point with a meaningful error message:

```typescript
const CANARY_KEY = '__canary__';
const CANARY_VALUE = 'midnight-verified';

async function unlockWithVerification(provider) {
  try {
    const existing = await provider.get(CANARY_KEY);
    if (existing === null) {
      // First unlock: write canary
      await provider.set(CANARY_KEY, CANARY_VALUE);
    }
    // If get() resolved without throwing, AES-GCM auth tag passed —
    // password is correct.
  } catch {
    throw new Error('Incorrect password — private state could not be decrypted');
  }
}
```

### Resetting private state

No built-in reset method exists in the official `PrivateStateProvider` interface.
If you need to clear all private state for an account (e.g. after a failed unlock),
implement a `resetStorage()` method in your provider wrapper that clears the
underlying storage for that `accountId`.

## Tested in

- `@midnight-ntwrk/midnight-js-level-private-state-provider@4.1.1`
- Midnight Mainnet
- Chrome + 1AM Wallet
- Reference dApp: https://midnight-private-auction.vercel.app

### DIFF
--- a/docs/guides/configure-providers.mdx
+++ b/docs/guides/configure-providers.mdx
@@ -331,10 +331,11 @@ The `deployContract` and `findDeployedContract` functions take the `MidnightProv
 For more information on using these functions, see the [Midnight.js SDK](/sdks/official/midnight-js) documentation.
 
 :::
-## Using levelPrivateStateProvider in browser dApps
+
+## Using levelPrivateStateProvider in browser DApps
 
 The `levelPrivateStateProvider` works in browser environments. Set `cryptoBackend: 'webcrypto'`
-to make the backend selection explicit — the SDK auto-detects Web Crypto availability, but
+to make the backend selection explicit. The SDK auto-detects Web Crypto availability, but
 setting it explicitly ensures predictable behavior across environments:
 
 ```typescript
@@ -345,7 +346,7 @@ const providers: MyProviders = {
     privateStateStoreName: 'my-contract-private-state',
     signingKeyStoreName: 'my-contract-private-state-signing-keys',
     privateStoragePasswordProvider: () => 'your-encryption-password',
-    cryptoBackend: 'webcrypto', // explicit — recommended for browser dApps
+    cryptoBackend: 'webcrypto', // explicit; recommended for browser DApps
   }),
   // ...other providers
 };
@@ -357,7 +358,7 @@ browsers, so no additional configuration is needed for storage.
 ### Early password validation with a canary
 
 Without additional checks, a wrong password isn't caught until the app attempts to decrypt
-real private state data — which may happen deep inside a transaction flow — producing an
+real private state data, which may happen deep inside a transaction flow, producing an
 opaque WebCrypto `OperationError` instead of a clear message at unlock time.
 
 A canary value written on first unlock lets you validate the password immediately:
@@ -373,10 +374,10 @@ async function unlockWithVerification(provider: PrivateStateProvider) {
       // First unlock: write canary
       await provider.set(CANARY_KEY, CANARY_VALUE);
     }
-    // If get() resolved without throwing, AES-GCM auth tag passed —
+    // If get() resolved without throwing, AES-GCM auth tag passed.
     // password is correct.
   } catch {
-    throw new Error('Incorrect password — private state could not be decrypted');
+    throw new Error('Incorrect password: private state could not be decrypted');
   }
 }
 ```
--- a/docs/guides/configure-providers.mdx
+++ b/docs/guides/configure-providers.mdx
@@ -345,6 +345,7 @@ const providers: MyProviders = {
   privateStateProvider: levelPrivateStateProvider({
     privateStateStoreName: 'my-contract-private-state',
     signingKeyStoreName: 'my-contract-private-state-signing-keys',
+    accountId: walletAddress,
     privateStoragePasswordProvider: () => 'your-encryption-password',
     cryptoBackend: 'webcrypto', // explicit; recommended for browser DApps
   }),
@@ -381,6 +382,7 @@ async function unlockWithVerification(provider: PrivateStateProvider) {
   }
 }
 ```
+
 ## Next steps
 
 With your providers configured, you are ready to deploy and interact with smart contracts on the Midnight Network. The following resources cover the next steps in that workflow:
--- a/docs/guides/configure-providers.mdx
+++ b/docs/guides/configure-providers.mdx
@@ -331,7 +331,55 @@ The `deployContract` and `findDeployedContract` functions take the `MidnightProv
 For more information on using these functions, see the [Midnight.js SDK](/sdks/official/midnight-js) documentation.
 
 :::
+## Using levelPrivateStateProvider in browser dApps
 
+The `levelPrivateStateProvider` works in browser environments. Set `cryptoBackend: 'webcrypto'`
+to make the backend selection explicit — the SDK auto-detects Web Crypto availability, but
+setting it explicitly ensures predictable behavior across environments:
+
+```typescript
+import { levelPrivateStateProvider } from '@midnight-ntwrk/midnight-js-level-private-state-provider';
+
+const providers: MyProviders = {
+  privateStateProvider: levelPrivateStateProvider({
+    privateStateStoreName: 'my-contract-private-state',
+    signingKeyStoreName: 'my-contract-private-state-signing-keys',
+    privateStoragePasswordProvider: () => 'your-encryption-password',
+    cryptoBackend: 'webcrypto', // explicit — recommended for browser dApps
+  }),
+  // ...other providers
+};
+```
+
+The `level` package automatically selects `browser-level` (IndexedDB) when bundled for
+browsers, so no additional configuration is needed for storage.
+
+### Early password validation with a canary
+
+Without additional checks, a wrong password isn't caught until the app attempts to decrypt
+real private state data — which may happen deep inside a transaction flow — producing an
+opaque WebCrypto `OperationError` instead of a clear message at unlock time.
+
+A canary value written on first unlock lets you validate the password immediately:
+
+```typescript
+const CANARY_KEY = '__canary__';
+const CANARY_VALUE = 'midnight-verified';
+
+async function unlockWithVerification(provider: PrivateStateProvider) {
+  try {
+    const existing = await provider.get(CANARY_KEY);
+    if (existing === null) {
+      // First unlock: write canary
+      await provider.set(CANARY_KEY, CANARY_VALUE);
+    }
+    // If get() resolved without throwing, AES-GCM auth tag passed —
+    // password is correct.
+  } catch {
+    throw new Error('Incorrect password — private state could not be decrypted');
+  }
+}
+```
 ## Next steps
 
 With your providers configured, you are ready to deploy and interact with smart contracts on the Midnight Network. The following resources cover the next steps in that workflow:


docs/guides/configure-providers.mdx
```diff
@@ -331,10 +331,58 @@ The `deployContract` and `findDeployedContract` functions take the `MidnightProv
 For more information on using these functions, see the [Midnight.js SDK](/sdks/official/midnight-js) documentation.
 
 :::
+## Using levelPrivateStateProvider in browser dApps
```
oduameh: Style: use "DApps" (capital D) per project convention. Also, insert a blank line between the admonition close `:::` and this heading to prevent MDX from parsing the heading as part of the admonition block.

```suggestion

## Using levelPrivateStateProvider in browser DApps
```

docs/guides/configure-providers.mdx
```diff
@@ -331,10 +331,58 @@ The `deployContract` and `findDeployedContract` functions take the `MidnightProv
...
 
+The `levelPrivateStateProvider` works in browser environments. Set `cryptoBackend: 'webcrypto'`
+to make the backend selection explicit — the SDK auto-detects Web Crypto availability, but
```
oduameh: Style: the project house style does not use em-dashes. Replace with a period to split into two sentences.

```suggestion
to make the backend selection explicit. The SDK auto-detects Web Crypto availability, but
```

docs/guides/configure-providers.mdx
```diff
@@ -331,10 +331,58 @@ The `deployContract` and `findDeployedContract` functions take the `MidnightProv
...
+setting it explicitly ensures predictable behavior across environments:
+
+```typescript
+import { levelPrivateStateProvider } from '@midnight-ntwrk/midnight-js-level-private-state-provider';
+
+const providers: MyProviders = {
+  privateStateProvider: levelPrivateStateProvider({
```
oduameh, line 345: Technical: the v4.0.4 API reference shows `accountId` as a **required** parameter in `LevelPrivateStateProviderConfig` (alongside `privateStoragePasswordProvider`). The function signature is:

```
Partial<LevelPrivateStateProviderConfig>
  & Pick<LevelPrivateStateProviderConfig, "privateStoragePasswordProvider" | "accountId">
```

The existing code block earlier on this page (line ~97) has the same gap, so this may need a coordinated fix across both examples. Consider adding `accountId: walletAddress,` to match the current API.

See the [LevelPrivateStateProviderConfig](https://docs.midnight.network/api-reference/midnight-js/@midnight-ntwrk/midnight-js-level-private-state-provider/interfaces/LevelPrivateStateProviderConfig) reference.

docs/guides/configure-providers.mdx
```diff
@@ -331,10 +331,58 @@ The `deployContract` and `findDeployedContract` functions take the `MidnightProv
...
+    privateStateStoreName: 'my-contract-private-state',
+    signingKeyStoreName: 'my-contract-private-state-signing-keys',
+    privateStoragePasswordProvider: () => 'your-encryption-password',
+    cryptoBackend: 'webcrypto', // explicit — recommended for browser dApps
```
oduameh: Style: replace the em-dash with a semicolon and use "DApps" (capital D).

```suggestion
    cryptoBackend: 'webcrypto', // explicit; recommended for browser DApps
```

docs/guides/configure-providers.mdx
```diff
@@ -331,10 +331,58 @@ The `deployContract` and `findDeployedContract` functions take the `MidnightProv
...
+};
+```
+
+The `level` package automatically selects `browser-level` (IndexedDB) when bundled for
+browsers, so no additional configuration is needed for storage.
+
+### Early password validation with a canary
+
+Without additional checks, a wrong password isn't caught until the app attempts to decrypt
+real private state data — which may happen deep inside a transaction flow — producing an
```
oduameh: Style: replace em-dashes with commas.

```suggestion
real private state data, which may happen deep inside a transaction flow, producing an
```

docs/guides/configure-providers.mdx
```diff
@@ -331,10 +331,58 @@ The `deployContract` and `findDeployedContract` functions take the `MidnightProv
...
+const CANARY_VALUE = 'midnight-verified';
+
+async function unlockWithVerification(provider: PrivateStateProvider) {
+  try {
+    const existing = await provider.get(CANARY_KEY);
+    if (existing === null) {
+      // First unlock: write canary
+      await provider.set(CANARY_KEY, CANARY_VALUE);
+    }
+    // If get() resolved without throwing, AES-GCM auth tag passed —
```
oduameh: Style: replace em-dash with a period.

```suggestion
    // If get() resolved without throwing, AES-GCM auth tag passed.
```

docs/guides/configure-providers.mdx
```diff
@@ -331,10 +331,58 @@ The `deployContract` and `findDeployedContract` functions take the `MidnightProv
...
+    // password is correct.
+  } catch {
+    throw new Error('Incorrect password — private state could not be decrypted');
```
oduameh: Style: replace em-dash with a colon.

```suggestion
    throw new Error('Incorrect password: private state could not be decrypted');
```

docs/guides/configure-providers.mdx
```diff
@@ -331,10 +331,58 @@ The `deployContract` and `findDeployedContract` functions take the `MidnightProv
...
+  }
+}
+```
```
oduameh, line 384: Formatting: add a blank line between this closing code fence and the `## Next steps` heading that follows it. Markdown parsers require blank lines before headings for reliable rendering.

oduameh (changes requested):
Thanks for this contribution, @pplmaverick. The browser-specific guidance for `levelPrivateStateProvider` and the canary password validation pattern are practical additions that fill a real documentation gap.

I found a few items that need attention before merging:

**Style fixes** (6 batch-committable suggestions)
- Replace `dApps` with `DApps` (capital D) per project convention (2 instances)
- Replace em-dashes with alternative punctuation; the docs house style does not use them (5 instances)

**Formatting** (1 comment)
- Insert blank lines around the new heading and before `## Next steps`

**Technical note** (1 comment)
- The v4.0.4 API reference shows `accountId` as a required parameter in `LevelPrivateStateProviderConfig`. See the inline comment for details.

The 6 suggestion comments are batch-committable from the GitHub UI.

pplmaverick:
All suggestions applied. Thank you for the review!

pplmaverick:
All comments resolved and suggestions applied. Ready for your re-review when you have a moment.

docs/guides/configure-providers.mdx
```diff
@@ -332,9 +332,60 @@ For more information on using these functions, see the [Midnight.js SDK](/sdks/o
 
 :::
 
+## Using levelPrivateStateProvider in browser DApps
+
+The `levelPrivateStateProvider` works in browser environments. Set `cryptoBackend: 'webcrypto'`
+to make the backend selection explicit. The SDK auto-detects Web Crypto availability, but
+setting it explicitly ensures predictable behavior across environments:
```
oduameh, line 339: Setting this explicitly is riskier than leaving it unset. `resolveCryptoBackend` throws when `'webcrypto'` is requested but `crypto.subtle` is missing, whereas omitting it falls back to the Noble backend and keeps working.

```suggestion
The `levelPrivateStateProvider` works in browser environments. The SDK resolves the crypto
backend automatically: it uses Web Crypto where `crypto.subtle` is available and falls back to
a pure-JS implementation where it is not. Leaving `cryptoBackend` unset is the safer default,
because requesting `'webcrypto'` explicitly turns an unavailable Web Crypto API into a thrown
error rather than a fallback. That matters in non-secure contexts, such as a page served over
plain `http://` on a non-loopback host, and in React Native:
```

docs/guides/configure-providers.mdx
```diff
@@ -332,9 +332,60 @@ For more information on using these functions, see the [Midnight.js SDK](/sdks/o
...
+```typescript
+import { levelPrivateStateProvider } from '@midnight-ntwrk/midnight-js-level-private-state-provider';
+
+const providers: MyProviders = {
+  privateStateProvider: levelPrivateStateProvider({
+    privateStateStoreName: 'my-contract-private-state',
+    signingKeyStoreName: 'my-contract-private-state-signing-keys',
+    accountId: walletAddress,
+    privateStoragePasswordProvider: () => 'your-encryption-password',
+    cryptoBackend: 'webcrypto', // explicit; recommended for browser DApps
```
oduameh, line 350: Following from the note above, so the example matches the guidance. The error from an explicit `'webcrypto'` on an unsupported runtime surfaces on the first `get`/`set` rather than here, which is worth flagging to anyone who does opt in.

```suggestion
    // cryptoBackend: 'webcrypto', // opt in only if you require Web Crypto and run under HTTPS or localhost
```

docs/guides/configure-providers.mdx
```diff
@@ -332,9 +332,60 @@ For more information on using these functions, see the [Midnight.js SDK](/sdks/o
...
+  }),
+  // ...other providers
+};
+```
+
+The `level` package automatically selects `browser-level` (IndexedDB) when bundled for
+browsers, so no additional configuration is needed for storage.
```
oduameh, line 357: The substitution is real, via `level`'s package.json `browser` field (there is no `exports` map), but it only applies to genuinely browser-targeted bundles. Next.js server components, API routes, and Edge functions resolve to the native `classic-level` instead, which will not run there. Worth scoping, since a Vercel-hosted DApp is a likely reader.

```suggestion
The `level` package resolves to `browser-level` (IndexedDB) in browser bundles through its
package.json `browser` field, which Vite and webpack honor automatically, so no extra storage
configuration is needed. Code that runs outside a browser bundle, such as Next.js server
components, API routes, or Edge functions, resolves to the native `classic-level` instead, so
keep private state access on the client.
```

docs/guides/configure-providers.mdx
```diff
@@ -332,9 +332,60 @@ For more information on using these functions, see the [Midnight.js SDK](/sdks/o
...
+
+### Early password validation with a canary
+
+Without additional checks, a wrong password isn't caught until the app attempts to decrypt
+real private state data, which may happen deep inside a transaction flow, producing an
+opaque WebCrypto `OperationError` instead of a clear message at unlock time.
```
oduameh, line 363: `OperationError` is specific to the Web Crypto backend. The Noble backend throws a plain `Error` with different wording, so a reader matching on the error name would miss it.

```suggestion
Without additional checks, a wrong password isn't caught until the app attempts to decrypt
real private state data, which may happen deep inside a transaction flow. The failure also
looks different depending on the active backend: Web Crypto rejects with a `DOMException`
named `OperationError`, while the pure-JS backend throws a plain `Error`.
```

docs/guides/configure-providers.mdx
```diff
@@ -332,9 +332,60 @@ For more information on using these functions, see the [Midnight.js SDK](/sdks/o
...
+      // First unlock: write canary
+      await provider.set(CANARY_KEY, CANARY_VALUE);
+    }
+    // If get() resolved without throwing, AES-GCM auth tag passed.
+    // password is correct.
+  } catch {
+    throw new Error('Incorrect password: private state could not be decrypted');
+  }
+}
+```
```
oduameh, line 384: This is the main change. As written, a `null` read is treated as first unlock, but `get()` also returns `null` for a wrong password whenever the canary key was never written, so a wrong password gets written in as the canary.

Splitting provisioning from verification removes that path. This version also adds the missing `PrivateStateProvider` import, preserves the underlying error via `cause` rather than discarding it, and distinguishes a missing canary from a bad password.

````suggestion
```typescript
import type { PrivateStateProvider } from '@midnight-ntwrk/midnight-js-types';

const CANARY_KEY = '__canary__';
const CANARY_VALUE = 'midnight-verified';

// Call once, when the account's store is created, using a password you know is correct.
// Never call this to repair a missing canary: at that point a missing canary and a wrong
// password look identical, and you would be storing the wrong password as the reference.
async function provisionCanary(provider: PrivateStateProvider): Promise<void> {
  await provider.set(CANARY_KEY, CANARY_VALUE);
}

// Call on every later unlock.
async function verifyPassword(provider: PrivateStateProvider): Promise<void> {
  let stored: unknown;
  try {
    stored = await provider.get(CANARY_KEY);
  } catch (cause) {
    // Decryption failed, so the AES-GCM authentication tag did not verify.
    // Other faults reach here too, such as an unavailable crypto backend,
    // so keep the original error attached.
    throw new Error('Incorrect password: private state could not be decrypted', { cause });
  }

  if (stored === null) {
    throw new Error('No canary stored for this account. Provision one when the store is created.');
  }

  if (stored !== CANARY_VALUE) {
    throw new Error('Canary mismatch: private state may be corrupt.');
  }
}
```
````

**Review by oduameh — CHANGES_REQUESTED** [review]
Thanks for the quick turnaround on the last round, the style points are all applied cleanly.

I took this one deeper on the technical side, verifying against `@midnight-ntwrk/midnight-js@4.1.1` and the `midnight-js` source at tag `v4.1.1`. The browser configuration half checks out well: `cryptoBackend` exists and is exactly `'webcrypto' | 'noble'`, `accountId` is required and a wallet address is the documented value, and the config snippet compiles both with and without generics. Your point about `verifyPassword()` being only an in-memory PBKDF2 cache-consistency check is also correct (`storage-encryption.ts` L142-145).

Two things need changing before this can land, both in code readers will copy.

**1. The canary can silently accept a wrong password.**

`subLevelMaybeGet` swallows only `error.code === 'LEVEL_NOT_FOUND'` into `null` and rethrows everything else. So `get('__canary__')` returns `null` without throwing whenever that key was never written, no matter whether the password is right or wrong.

That is exactly the retrofit case: an account that already holds real private state under the correct password, with no canary yet. A wrong password there produces `null`, the snippet takes its "first unlock" branch, and writes the canary **under the wrong password**. From then on the canary validates the wrong password while the real private state stays encrypted under the original one. The check ends up certifying the failure it was added to catch.

The pattern works, but only if the canary is provisioned exactly once alongside store creation under a password already known to be correct. It cannot be bootstrapped from a `null` read. Suggestion below splits provisioning from verification.

**2. Recommending explicit `cryptoBackend: 'webcrypto'` makes some browsers fail that currently work.**

`resolveCryptoBackend` validates an explicit preference:

```ts
if (preference === 'webcrypto') {
  if (!isWebCryptoAvailable()) {
    throw new Error("Web Crypto API is not available. Use the 'noble' crypto backend or run in a secure context (HTTPS or localhost).");
  }
```

Omit the option and it falls back to the Noble backend and works. Set it explicitly and the same runtime condition becomes a thrown error. `isWebCryptoAvailable()` is feature detection rather than browser detection, so it returns false in a non-secure context (plain `http://` on a non-loopback host) and in React Native. The failure also surfaces as an async rejection on the first `get`/`set`, not at the `levelPrivateStateProvider(...)` call, so it is easy to misdiagnose.

Worth noting the two interact: that "Web Crypto API is not available" error is thrown inside `get()`, which the current canary wraps in a bare `catch {}` that reports everything as `'Incorrect password'`. A developer with a perfectly good password on `http://` would be told their password is wrong.

**Smaller points, all in suggestions below:** `OperationError` is backend-specific (Web Crypto rejects with a `DOMException` named `OperationError`; the Noble backend throws a plain `Error`), the bundler claim needs scoping to client bundles, and the canary snippet needs its `PrivateStateProvider` import.

**Two things I could not leave as inline suggestions, since they sit outside this diff:**

- The parameter table at lines 106-110 still lists only three parameters. `accountId` is now required and `cryptoBackend` is missing, and the earlier example at line 97 omits `accountId` too. Since this PR is what introduces both to the page, it would be good to bring that table into line here.
- The canary snippet type-checks only because bare `PrivateStateProvider` defaults its state type to `any`. Everywhere else this page writes `levelPrivateStateProvider<PrivateStateId, MyPrivateState>(...)`, and under that typing `set('__canary__', 'midnight-verified')` will not compile, because a string is not `MyPrivateState`. Related: `PrivateStateId` is just `string`, so the canary shares a keyspace with real private state and will show up in `exportPrivateStates()`. A dedicated store for the canary would avoid both. Happy to talk through options if you would rather solve it a different way.

**Finally, two corrections to the PR description rather than the diff:**

- Browser compatibility came from midnight-js #798 ("feat: web crypto migration"), not #827. #827 added the `CryptoBackend` abstraction, the Noble fallback, and `isWebCryptoAvailable()` itself, and its own description says "PR #798 ... enabling browser support."
- "No built-in reset method exists in the official `PrivateStateProvider` interface" is not right: `clear()`, `remove()`, and `clearSigningKeys()` all exist on it. Worth knowing before the proposed `resetStorage()` section gets written.

One process note: only the Vercel and CLA checks ran here, so this PR has had no `build` or `vale` verification. That looks related to the head branch being the fork's `main`.

oduameh:
Heads-up on an upcoming restructure that affects this PR's target file. As part of the guides consolidation (#1166, #1179, #1218, and #1220 are the earlier slices), `docs/guides/configure-providers.mdx` is being merged into a single deploy-and-operate assembly together with `deploy-mn-app`, `query-the-blockchain`, and the updatability procedures. The standalone page will be deleted, with a permanent redirect pointing its URL at the new page's providers section.

Rather than lose this PR's content, the browser `levelPrivateStateProvider` procedure would fit naturally as its own module in the new assembly. Happy to help re-target the diff once the assembly PR lands, or to fold it in as a follow-up if this merges first, whichever is easier on your end.

pplmaverick:
Thank you for the thorough technical review — this is exactly the level of scrutiny this kind of documentation needs.

I've gone through both rounds of feedback and understand the issues:

**Canary logic:** The split between `provisionCanary()` and `verifyPassword()` is the right design. The original approach of bootstrapping from a null read is fundamentally unsound — a missing canary and a wrong password are indistinguishable at that point, so you'd end up certifying the failure you were trying to catch. I'll apply your suggested change as-is.

**`cryptoBackend: 'webcrypto'`:** Understood. Explicit opt-in converts an unavailable Web Crypto API from a graceful Noble fallback into a thrown error, which is strictly worse for most DApp deployments. I'll follow your suggested wording and leave it unset by default.

I'll also apply the remaining inline suggestions (OperationError scoping, Next.js server component caveat, `PrivateStateProvider` import) and correct the PR description (`#827` → `#798`, and remove the "no built-in reset" claim given `clear()` / `remove()` / `clearSigningKeys()` exist).

On the restructure: since #1221 has already merged, I'd like to re-target this content at the new `deploy-and-operate` assembly rather than have it get lost. Would you be able to point me to where the browser `levelPrivateStateProvider` section would fit in the new file, or would it be easier if you folded it in as a follow-up? Happy to go either way — just want to make sure the content lands.

oduameh:
**Placement in the new assembly.** `docs/guides/configure-providers.mdx` no longer exists on `main`; it was merged into `docs/guides/deploy-and-operate.mdx` (the old URL redirects there). Two homes for this content:

- The browser configuration notes (leave `cryptoBackend` unset so the Noble fallback stays available, `level` resolves to `browser-level`/IndexedDB in client bundles) go as a short note directly after step 3 of "Configuring providers for a contract", which is where `levelPrivateStateProvider` is created.
- The canary password check is its own procedure. Add it as a new `###` section under "Provider implementations", not inside the main procedure.

**Branch.** Open the re-targeted change from a feature branch off current `main`, not the fork's `main`. The current head is why `build` and `vale` never ran on this PR, and the branch now conflicts with the deletion.

**Carry over the five open round-2 suggestions** (canary provisioning split, `cryptoBackend` wording and example, client-bundle scoping, `OperationError` scoping, `PrivateStateProvider` import). Re-checked against `@midnight-ntwrk/midnight-js-level-private-state-provider@4.1.1` (still `latest`) and `5.0.0-beta.6`: the explicit-`webcrypto` throw, the `null`-on-missing-key read, and the cache-only `verifyPassword()` are unchanged, and no reset or password-verify API has been added. All five still apply as written.

Leaving this in changes requested until the re-targeted branch is up.

pplmaverick:
Reopened as #1257, re-targeted at `docs/guides/deploy-and-operate.mdx` following the guides consolidation in #1221.

All five round-2 suggestions from your last review are carried over: canary provisioning split, `cryptoBackend` wording, client-bundle scoping, `OperationError` scoping, and `PrivateStateProvider` import. The new PR is opened from a feature branch off current `main`, so build and vale checks should run this time.

Closing this PR.

pplmaverick:
closed.

PR #1102, opened 2026-06-25
docs(guides): expand Windows setup with verification, troubleshooting, and quick-reference

notsoblack:
## Summary

Significantly improves the `windows-compact-setup.mdx` guide to reduce friction for Windows developers onboarding to Midnight.

### Added
- **Prerequisites table** — Windows version requirements, virtualization check, disk space, and admin rights
- **Virtualization verification** — PowerShell `systeminfo` command to confirm Intel VT-x / AMD-V is enabled before starting
- **Step-by-step verification** after every major install (WSL, Docker, proof server, Compact compiler)
  - Example: `wsl -l -v` should show VERSION 2, `docker run hello-world`, `curl http://localhost:6300`
- **Detached mode for proof server** — `docker run -d --restart unless-stopped ...` so developers don't need to keep a terminal window open permanently
- **Proof server health check** — `curl http://localhost:6300` to confirm connectivity before proceeding
- **Dynamic VSIX download instructions** — Replaced hardcoded `compact-0.2.13` link with guidance to always fetch the latest release
- **Comprehensive troubleshooting tables** covering:
  - WSL: elevation errors, kernel update, virtualization disabled
  - Docker: command not found in WSL, daemon not running, port conflicts
  - Proof server: firewall blocking, container exiting immediately
  - Compact: PATH not updated, DNS failures, VSIX architecture mismatch
- **Quick-reference checklist** — A printable checklist at the end so developers can verify every step before moving to the hello-world tutorial

### Improved
- Structure now follows "install → verify → troubleshoot" for each component
- Added expected output examples so developers know when a step succeeded
- Clarified that Compact must be installed inside WSL, not Windows natively

## Why this matters

The current guide assumes everything works smoothly, but Windows + WSL + Docker is a common failure point for new developers. These additions reduce support burden and improve first-time setup success rate.

## Checklist
- [x] Content follows the [Midnight documentation style guide](https://docs.midnight.network/)
- [x] Headings are maximum three levels deep
- [x] No private repository references
- [x] Language is precise and focused on architecture/processes

PR #1132, opened 2026-07-15
Expand Windows Compact setup guide (supersedes #1102)

oduameh:
## Summary

Supersedes #1102 by @notsoblack. Takes the community contribution and applies all review feedback from @oduameh.

Credit: The original content (prerequisites table, step-by-step verification, detached-mode proof server, troubleshooting tables, quick-reference checklist) was contributed by @notsoblack in #1102.

### DIFF
--- a/docs/guides/windows-compact-setup.mdx
+++ b/docs/guides/windows-compact-setup.mdx
@@ -5,13 +5,13 @@ title: Windows Compact setup
 description: Learn how to set up the Midnight development environment on Windows using Windows Subsystem for Linux (WSL) with Docker and the Compact compiler.
 sidebar_position: 60
 sidebar_label: "Windows Compact setup"
-tags: [installation, windows, WSL, compact, setup]
-toc_max_heading_level: 2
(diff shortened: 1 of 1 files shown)


nstanford5 (changes requested):
I am partial to not writing guides for software we don't control. Most of this guide is a duplicate of what you find in the WSL install instructions and now we need to maintain parity with an outside source. And then it duplicates the install instructions from getting-started, which creates another place to maintain these.

We should only link out to the WSL install instructions and list the results expected. For example:

- WSL2
- Ubuntu LTS

"Then follow the standard install instructions here(link to getting-started"

We should not duplicate instructions, instead leaning on one source of truth that can be pointed at from different parts of the process.

nstanford5:
If you are scrapping this page in https://github.com/midnightntwrk/midnight-docs/pull/1134 -- my vote is to scrap it all together.

oduameh:
> If you are scrapping this page in #1134 -- my vote is to scrap it all together.

I agree

PR #1147, opened 2026-07-23
docs: add DUST generation formula, worked example, and registration fee mechanics

wbaxterh:
## What this improves

The [DUST architecture](https://docs.midnight.network/docs/learn/introduction/native-token/) concepts page lists the initial DUST parameters and says the generation rate "depends on the amount of NIGHT held, the ratio, and time to cap" — but never connects them: there is no formula, no worked example, and no explanation of where the "~1 week generation time" comment on `generation_decay_rate` comes from. The registration section also compresses the most practically important mechanic (how a brand-new wallet pays for its own registration when it cannot hold DUST yet) into a single sentence about "backdating".

This PR:

1. **Adds a "Generation rate and capacity" section**: the DUST value formula (`min(N × generation_decay_rate × t, N × night_dust_ratio)`), the derivation showing time-to-cap is the same for any holding size (`5×10⁹ ÷ 8267 ≈ 604,800 s ≈ 7 days` — the origin of the "~1 week" comment), and a worked example for 100 NIGHT (≈71 DUST/day, capping at 500 DUST).
2. **Expands "Registrations and fees"**: the bootstrap problem, retroactive generation and the `allow_fee_payment` allowance, the two constraints on the self-funded path (only guaranteed-section NIGHT inputs count; the allowance must not exceed accrued DUST, otherwise `InsufficientDustForRegistrationFee`), and why registration transactions self-spend the wallet's NIGHT.
3. **Repairs a truncated sentence** in the network usage section ("It function" → completed list introduction).

## Accuracy

Formulas, parameter semantics, and registration behaviour are verified against the ledger implementation (`dust.rs`: `generationless_fee_availability`, `DustParameters`) — this mechanic is the subject of midnightntwrk/midnight-node#1896 / midnightntwrk/midnight-node#1903, where I worked through it at the source level. The worked-example numbers are computed from the initial parameters already documented on the page.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nstanford5:
Thanks for the submission @wbaxterh -- @oduameh please review

docs/concepts/dust-architecture.mdx
```diff
@@ -84,6 +84,23 @@ flowchart TD
 
 The rate of generation depends on the amount of NIGHT held ($N$), the ratio of the DUST cap to NIGHT held ($\rho$), and "time to cap" ($\Delta$).
 
+#### Generation rate and capacity
+
+Two ledger parameters govern how fast DUST accrues and how much can accumulate (their initial values appear in [Initial DUST parameters](#initial-dust-parameters) below):
+
+- `generation_decay_rate`: the Specks of DUST generated per Star of backing NIGHT per second.
```
oduameh: The same field sets the decay slope at the same linear rate once the backing NIGHT is spent (`ledger/tests/dust.rs:161-165` shows the balance falling linearly to exactly zero). Worth saying here, since the page refers to a decay rate under **Spending rules** without ever connecting it to this parameter.

```suggestion
- `generation_decay_rate`: the Specks of DUST generated per Star of backing NIGHT per second. The same rate applies in reverse after the backing NIGHT is spent, which is where the "decay" in the name comes from.
```

docs/concepts/dust-architecture.mdx
```diff
@@ -84,6 +84,23 @@ flowchart TD
...
+- `night_dust_ratio`: the maximum Specks of DUST per Star of backing NIGHT.
+
+For a NIGHT UTXO of value $N$ Stars, at $t$ seconds after its creation, the associated DUST value in Specks is the smaller of $N \times g \times t$ and $N \times \rho$, where $g$ is `generation_decay_rate` and $\rho$ is `night_dust_ratio`.
```
oduameh: This reads as the value function for all time, but it is phase one only: it holds while the backing NIGHT is unspent, and assumes the DUST UTXO's initial value is zero. `DustOutput::updated_value` (`ledger/src/dust.rs:281-330`) has two further phases, and **DUST value and spends** below already tells the reader there are four segments.

```suggestion
While the backing NIGHT UTXO remains unspent, the associated DUST value in Specks at $t$ seconds after the UTXO's creation is the smaller of $N \times g \times t$ and $N \times \rho$, where $N$ is the UTXO's value in Stars, $g$ is `generation_decay_rate`, and $\rho$ is `night_dust_ratio`.
```

docs/concepts/dust-architecture.mdx
```diff
@@ -84,6 +84,23 @@ flowchart TD
...
+
+Because both the rate and the cap scale linearly with $N$, the time to reach capacity is the same for any amount of NIGHT: $\Delta = \rho \div g = 5 \times 10^9 \div 8267 \approx 604800$ seconds, which is where the "about one week" generation time comes from.
+
+For example, suppose you hold a single NIGHT UTXO of 100 NIGHT ($10^8$ Stars) with a registered DUST address:
+
+- Your DUST balance grows by about 830 billion Specks per second — roughly 0.00083 DUST per second, or 71 DUST per day.
```
oduameh: $10^8 \times 8267$ is 826.7 billion, so 830 rounds up by 0.4% and disagrees with the `0.00083` beside it, which is rounded correctly. Also drops the em-dash.

```suggestion
- Your DUST balance grows by about 827 billion Specks per second, roughly 0.00083 DUST per second, or 71 DUST per day.
```

docs/concepts/dust-architecture.mdx
```diff
@@ -84,6 +84,23 @@ flowchart TD
...
+- It reaches its cap after about a week, at 500 DUST in total — 5 DUST per NIGHT.
```
oduameh:
```suggestion
- It reaches its cap after about a week, at 500 DUST in total, or 5 DUST per NIGHT.
```

oduameh (commented):
Thanks Wes, this is a strong addition and the page needed it.

I verified the formula, the parameter units, the worked example, and the registration mechanics against `midnight-ledger` @ `272c25f` (branch `ledger-8`). Everything checks out, including `allow_fee_payment`, the guaranteed-section restriction, and the exact spelling of `InsufficientDustForRegistrationFee`. The worked-example numbers recompute correctly. `time_to_cap()` is literally `night_dust_ratio.div_ceil(generation_decay_rate)`, so the N-independence argument holds as written.

The suggestions below are about scope rather than accuracy. Two are worth taking:

- The formula reads as the complete value function, but it covers the generation phase only (and assumes an initial value of zero). "DUST value and spends" further down says the value has four linear segments, so as written the two sections disagree.
- `generation_decay_rate` also sets the decay slope, at the same linear rate. One clause covers it, and it explains the "decay" in the parameter name, which the page otherwise leaves unexplained.

The rest are the em-dashes, a rounding nit, and a link for "guaranteed section", which is not defined anywhere on this page.

`build` and `vale` pass. The `Vercel`, `deploy-preview`, and `link-check` failures all come from the PR being on a fork rather than upstream: the deploy cannot authorize and the "Comment broken links" step cannot write. Lychee itself passed, so there are no broken links. Pushing the branch to upstream would give you a working preview.

docs/concepts/dust-architecture.mdx
```diff
@@ -171,10 +188,16 @@ struct DustActions<S, P> {
 
 ### Registrations and fees
 
-`DustRegistration` links a NIGHT key to a DUST key.
+`DustRegistration` links a NIGHT key to a DUST key. Registrations happen sequentially.
+
+A first-time registration faces a bootstrap problem: the transaction itself costs DUST, but a wallet cannot hold DUST until it is registered. The ledger resolves this with *retroactive generation*. NIGHT UTXOs that are not yet generating DUST (their key has no previous registration) have already "virtually" accrued DUST under the same rate and cap formula as registered UTXOs, counted from each UTXO's creation time. A registration can declare a fee allowance (`allow_fee_payment`) to draw on that retroactive DUST and pay for its own gas — no pre-existing DUST or second wallet required.
```
oduameh: Three small things: the check is per-UTXO rather than per-key (`generationless_fee_availability` tests `!night_indices.contains_key(&i.initial_nonce())`, so it asks whether that particular UTXO is already generating), the bootstrap claim is about `DustRegistration` specifically, and the paragraph carries four dense sentences.

```suggestion
A first-time `DustRegistration` faces a bootstrap problem: the transaction itself costs DUST, but a wallet cannot hold DUST until its NIGHT key is registered. The ledger resolves this with *retroactive generation*.

A NIGHT UTXO that is not yet generating DUST has still "virtually" accrued DUST, under the same rate and cap formula as a generating UTXO, counted from that UTXO's own creation time. A registration can declare a fee allowance (`allow_fee_payment`) to draw on that retroactive DUST and pay for its own gas. No pre-existing DUST or second wallet is required.
```

docs/concepts/dust-architecture.mdx
```diff
@@ -171,10 +188,16 @@ struct DustActions<S, P> {
...
+
+Two constraints apply to the self-funded path:
+
+- Only the not-yet-generating NIGHT UTXOs *spent in the registration transaction's guaranteed section* count toward the retroactive allowance, and the declared allowance must not exceed what those inputs have accrued. A transaction that requests more fails validation with `InsufficientDustForRegistrationFee`.
+- Because retroactive DUST accrues from each UTXO's creation time, a freshly received NIGHT UTXO has accrued almost nothing. If the retroactive DUST cannot yet cover the registration fee, wait for more to accrue or fund the fee from an already-registered wallet.
```
oduameh: "Two constraints" does not quite map onto the two bullets: the first holds both constraints, and the second is guidance rather than a constraint. Splitting them also gives "guaranteed section" somewhere to link, since the term is not defined on this page.

```suggestion
Two limits apply to the self-funded path:

- Only the not-yet-generating NIGHT UTXOs spent in the registration transaction's [guaranteed section](./how-midnight-works/semantics.mdx) count toward the retroactive allowance.
- The declared allowance must not exceed what those inputs have accrued. A transaction that requests more fails validation with `InsufficientDustForRegistrationFee`.

Because retroactive DUST accrues from each UTXO's creation time, a freshly received NIGHT UTXO has accrued almost nothing. If the retroactive DUST cannot yet cover the registration fee, wait for more to accrue or fund the fee from an already-registered wallet.
```

wbaxterh:
All six applied in b153d693.

The two structural ones were right:

- **Phase scoping:** the formula read as the whole value function, which put it at odds with the four segments documented further down. Now bounded to "while the backing NIGHT UTXO remains unspent", with N defined inline.
- **`generation_decay_rate` as decay slope:** the page used the parameter, named it, and never said where "decay" came from. One clause covers it.

On the rounding: 10^8 x 8267 = 826.7 billion, so 830 and the `0.00083` sitting next to it couldn't both be right. They come from the same number now.

Also confirmed `how-midnight-works/semantics` actually defines the guaranteed phase before linking it, and pulled the em dashes page-wide rather than just the lines you flagged.

`yarn build` green locally, no broken links on the rendered page. I can't push this branch upstream from a fork, so the preview would have to come from your side. Happy to keep iterating here.

oduameh (commented):
Thanks for turning all six around so quickly, and for the clear write-up on what you changed. I rebuilt the site on top of current `main` to check the result: the em-dashes are gone, the numbers are right, the phase scoping reads correctly, and the `guaranteed section` link I suggested resolves properly (`/concepts/how-midnight-works/semantics`). Vale is clean at 0 errors and readability actually improved.

One thing I should have caught in my first pass and didn't, and unfortunately it lands right on the part of this PR that matters most.

**The math on this page never renders.** `remark-math` and `rehype-katex` aren't dependencies of this repo, and `docusaurus.config.js` configures no remark or rehype plugins. I confirmed it in a local build and then on the live site, which today serves this literally:

> The rate of generation depends on the amount of NIGHT held ($N$), the ratio of the DUST cap to NIGHT held ($\rho$), and "time to cap" ($\Delta$).

Dollar signs and backslashes, exactly as written. That's pre-existing, but this PR takes the page from 5 LaTeX spans to 14, and the new ones carry the formula and the derivation:

```
the smaller of $N \times g \times t$ and $N \times \rho$
$\Delta = \rho \div g = 5 \times 10^9 \div 8267 \approx 604800$
```

So the two things this PR exists to add are currently the least readable text on the page. I also want to flag that my earlier "promote it to a display equation" idea would have made this worse rather than better, so I'm glad you left it.

`dust-architecture.mdx` is the only page in the docs using real LaTeX (every other `$...$` in `docs/` is a shell variable inside a code block), so pulling KaTeX into the site for one page seems like the wrong trade. The suggestions below switch to Unicode in backticks, which renders everywhere and lets the numbers carry thousands separators. Two of them also split sentences that were still running past 25 words.

I've included line 85 as well, even though it's pre-existing, because leaving it would make the page half-converted. Two more are outside this diff and can't take suggestions:

- **Line 52**: `up to a cap ($\rho$)` becomes `up to a cap (`ρ`)`. Worth noting that `ρ` is a ratio rather than a cap, so `up to its cap of `N × ρ`` would be more accurate if you want to fix the substance at the same time.
- **Line 277**: `sequence numbers ($0, 1, 2...$)` becomes `sequence numbers (`0, 1, 2, ...`)`.

Entirely your call whether the pre-existing three ride along here or go in a separate PR. Happy either way, and happy to open that follow-up myself if you'd rather keep this one tightly scoped.

One process note that isn't yours to fix: no CI has run on `b153d693` at all, so the `build` and `vale` results showing on this PR are from the earlier commit. It's the fork-branch setup. I verified the build locally instead, and it passes.

docs/concepts/dust-architecture.mdx
```diff
@@ -84,6 +84,23 @@ flowchart TD
 
 The rate of generation depends on the amount of NIGHT held ($N$), the ratio of the DUST cap to NIGHT held ($\rho$), and "time to cap" ($\Delta$).
```
oduameh: Pre-existing rather than something you introduced, but it's the first math the reader meets, so converting it keeps the page consistent with the section below.

```suggestion
The rate of generation depends on the amount of NIGHT held (`N`), the ratio of the DUST cap to NIGHT held (`ρ`), and "time to cap" (`Δ`).
```

docs/concepts/dust-architecture.mdx
```diff
@@ -84,6 +84,23 @@ flowchart TD
...
 
+#### Generation rate and capacity
+
+Two ledger parameters govern how fast DUST accrues and how much can accumulate (their initial values appear in [Initial DUST parameters](#initial-dust-parameters) below):
+
+- `generation_decay_rate`: the Specks of DUST generated per Star of backing NIGHT per second. The same rate applies in reverse after the backing NIGHT is spent, which is where the "decay" in the name comes from.
+- `night_dust_ratio`: the maximum Specks of DUST per Star of backing NIGHT.
+
+While the backing NIGHT UTXO remains unspent, the associated DUST value in Specks at $t$ seconds after the UTXO's creation is the smaller of $N \times g \times t$ and $N \times \rho$, where $N$ is the UTXO's value in Stars, $g$ is `generation_decay_rate`, and $\rho$ is `night_dust_ratio`.
```
oduameh: This is the sentence that suffers most from the missing math renderer. Also splits it in two, since it was still running to 42 words.

```suggestion
While the backing NIGHT UTXO remains unspent, the associated DUST value in Specks at `t` seconds after the UTXO's creation is the smaller of `N × g × t` and `N × ρ`. Here `N` is the UTXO's value in Stars, `g` is `generation_decay_rate`, and `ρ` is `night_dust_ratio`.
```

docs/concepts/dust-architecture.mdx
```diff
@@ -84,6 +84,23 @@ flowchart TD
...
+
+Because both the rate and the cap scale linearly with $N$, the time to reach capacity is the same for any amount of NIGHT: $\Delta = \rho \div g = 5 \times 10^9 \div 8267 \approx 604800$ seconds, which is where the "about one week" generation time comes from.
```
oduameh: Same change, plus thousands separators, which the LaTeX form couldn't carry. `5 × 10^9 ÷ 8267` is exact to 604,814.32, and the ledger's own `time_to_cap()` returns 604,815 via `div_ceil`, so `≈ 604,800` stays honest.

```suggestion
Because both the rate and the cap scale linearly with `N`, the time to reach capacity is the same for any amount of NIGHT: `Δ = ρ ÷ g = 5,000,000,000 ÷ 8,267 ≈ 604,800` seconds. That is where the "about one week" generation time comes from.
```

docs/concepts/dust-architecture.mdx
```diff
@@ -84,6 +84,23 @@ flowchart TD
...
+
+For example, suppose you hold a single NIGHT UTXO of 100 NIGHT ($10^8$ Stars) with a registered DUST address:
```
oduameh: `$10^8$` renders as raw source too. Spelling the number out also saves the reader converting it.

```suggestion
For example, suppose you hold a single NIGHT UTXO of 100 NIGHT (100,000,000 Stars) with a registered DUST address:
```

wbaxterh:
All of them applied in 08de5af0, including the three pre-existing lines (52, 85, 277), so the page converts fully rather than half.

- **Line 52:** took the substance fix, "up to its cap of `N × ρ`". You're right that ρ alone is the ratio, not the cap.
- **Zero `$` left in the file:** nothing on the page depends on a renderer the repo doesn't ship.

Good catch on the rendering. I verified the MDX compiles, which it turns out proves nothing about what the reader sees. Agreed on not pulling KaTeX in for one page; Unicode in backticks reads fine everywhere I checked.

Rebuilt locally on top of your branch update (thanks for that): the converted strings all appear in the rendered page and there are no LaTeX leftovers. Ready for another look whenever.

oduameh (approved):
Verified at 08de5af0. Zero `$` left in the source and no LaTeX leftovers in the rendered HTML; `N × g × t`, `N × ρ`, `Δ = ρ ÷ g = 5,000,000,000 ÷ 8,267 ≈ 604,800`, `100,000,000 Stars`, and `0, 1, 2, ...` all render correctly. No em-dashes, the guaranteed-section link resolves, no broken links from this page, `yarn build` green on current `main`, and Vale clean at 0 errors.

Thanks for taking the substance fix on line 52 as well, and for converting the pre-existing lines so the page ends up consistent rather than half done.

Every technical claim here was checked against the ledger source on the first pass: the value formula and its linear-with-cap behaviour in `DustOutput::updated_value`, `time_to_cap()` being ratio over rate with no N term, the units on `generation_decay_rate` and `night_dust_ratio`, `allow_fee_payment`, the guaranteed-section-only restriction in `generationless_fee_availability`, and the exact spelling of `InsufficientDustForRegistrationFee`. All confirmed.

One non-blocking nit for a future pass: line 52 uses `N × ρ` about thirty lines before either symbol is defined at line 85. It reads fine in context, so I would not hold the PR for it.

Approving.

PR #1166, opened 2026-07-31
docs: add security and best practices guide (modular assembly)

oduameh:


docs/guides/security-best-practices.mdx
```diff
@@ -0,0 +1,702 @@
...
+## What an observer can see
+
+Zero-knowledge proofs hide your witness data, but a transaction still reveals a great deal. Know the surface before you defend it.
+
+| What the observer sees | Visible on-chain? |
+|---|---|
+| Which exported circuit you called | Yes, the entry point is part of the transaction |
+| Which contract you called | Yes, the contract address is public |
+| Arguments to ledger operations (`Set` and `Map` keys and values, `Counter` amounts) | Yes |
+| Values you wrap in `disclose()` | Yes, by definition |
```
nstanford5: Not necessarily -- not all values wrapped in disclose end up being public. Only values that are wrapped in disclose and cross into a public domain (ledger write, return from exported circuit, contract to contract calls)

docs/guides/security-best-practices.mdx
```diff
@@ -0,0 +1,702 @@
...
+## Authenticate a caller
+
+The task: gate a circuit so only one specific caller can run it. The wrong way is to compare `ownPublicKey()`, which is a witness the prover controls. The right way derives an identity from a secret the caller must know.
+
+<StepsProvider>
+<Step>
+
+### Declare a secret witness
+
+The caller proves who they are by knowing a secret. Declare it as a witness so it stays in private state and never reaches the ledger.
```
nstanford5: Witness !== private state. There is nothing wrong with accessing private state through a witness function, but it is also just as easily passed in as a circuit argument while still maintaining privacy. Given that this is a best-practices guide, it should probably show both

docs/guides/security-best-practices.mdx
```diff
@@ -0,0 +1,702 @@
...
+`ownPublicKey()` is a witness. The prover chooses its return value, and the protocol does not check it against the wallet that signed the transaction. An `assert(ownPublicKey().bytes == owner)` compares two prover-controlled values, so an attacker reads the public `owner` and returns it from a modified frontend. `ownPublicKey()` is only safe when you route a value *to* the caller, as the [shielded token tutorial](../tokens/shielded-token) does, where lying only hurts the prover.
+
+:::
+
+</Step>
+<Step>
+
+### Store the commitment
+
+At setup, derive the caller's identity and store it as the owner. `persistentHash` is witness-derived, so the ledger write needs `disclose()`.
```
nstanford5: It doesn't need disclose because it is witness-derived, but because it is being written to a public domain

nstanford5 (changes requested):

docs/guides/security-best-practices.mdx
oduameh: Good catch, fixed.

docs/guides/security-best-practices.mdx
oduameh: Agreed. Added a "Witness or circuit argument" note to the authenticate procedure showing the secret passed as a circuit argument (private by default) alongside the witness form, with the tradeoff between them. I compiled and executed the argument variant to confirm it behaves identically (owner passes, forged key fails).

docs/guides/security-best-practices.mdx
oduameh: Right, the causality was off. That sentence is gone in the restructure; step 3 now states disclosure is required because "writing the derived identity to the ledger moves a private value into a public position", and the new on-chain visibility note states the same rule.

nstanford5 (approved):

Olanetsoft (approved):

PR #1188, opened 2026-08-06
test: single-file uses: SHA bump probe (do not merge)

chrisferry:
> AI-assisted (Claude); no bot:ai-assisted label exists in this repo.

Diagnostic probe for the sha_pinning_required run-suppression affecting #1165/#1187: bumps one `uses:` SHA in one file. If Actions stays silent here, any PR changing action references is suppressed under the policy. Will be closed after observation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


chrisferry:
Probe complete: confirmed that with sha_pinning_required enabled, any PR modifying a uses: action reference gets no workflow runs at all (no github-actions check suite), regardless of author. See #1187.

Issue #1172, opened 2026-08-03
~80 stale testnet-02 references across org repos, six months after the Preview transition


protzi:
`rpc.testnet-02.midnight.network` / `indexer.testnet-02.midnight.network` are NXDOMAIN (checked against 8.8.8.8 / 1.1.1.1 / 9.9.9.9) since the Preview transition announced 2026-01-27 — but roughly 80 references to `testnet-02` persist across the org's repos, including `midnight-node-docker`'s `README.md` and `.envrc`, and are still echoed by third-party RPC directories.

The migration guide (`how-to/migrate-from-testnet-02-to-preview.mdx`) exists and is good — the problem is the stale references around it, which send every new integrator to a dead endpoint first (we hit it twice, in two separate evaluation sessions).

A sweep replacing `testnet-02` endpoints with Preview/Preprod equivalents (or a pointer to the migration guide) would spare that detour. Filing under docs per the `fix-content` template's spirit; the affected files span several org repos.

PR #1194, opened 2026-08-09
docs: replace retired testnet-02 endpoints in validator runbook

subheeksh5599:
## What

Replaces the retired `testnet-02` endpoints in the validator runbook (`docs/nodes/_run-a-validator/`) with the current Preview equivalents, per #1172.

- **step-3.mdx** — four `curl` examples pointed at `rpc.testnet-02.midnight.network` (NXDOMAIN since the Preview transition); now use `rpc.preview.midnight.network`.
- **step-1.mdx** — the supported-environments table mapped Cardano `preview` to Midnight `testnet-02`; now maps to the Midnight `preview` environment.
- **index.mdx** — the "Public endpoints" section linked `ogmios.testnet-02.midnight.network`, which no longer resolves and has no Preview equivalent. The guide's own Docker Compose stack (step 2) runs Ogmios locally via the `cardano-ogmios` container, so the section now points there instead of advertising a dead public endpoint.

## Why

`testnet-02` was retired on 2026-01-27; `rpc.testnet-02.midnight.network` and `ogmios.testnet-02.midnight.network` return NXDOMAIN (verified against public resolvers). Every new validator following these steps hits a dead endpoint first. The migration guide already documents the testnet-02 → Preview transition; this removes the stale references around it.

## Scope notes

- Historical release notes (`docs/relnotes/`) and the versioned docs (`main_versioned_docs/`) keep their `testnet-02` mentions intentionally — they document what was true at the time.
- `step-3.mdx` line 50 still references `midnight-node-docker`'s `envs/testnet-02/pc-chain-config.json` because that external repo still ships that path; it was left untouched.

### DIFF
--- a/docs/nodes/_run-a-validator/index.mdx
+++ b/docs/nodes/_run-a-validator/index.mdx
@@ -169,9 +169,7 @@ An existing Cardano relay node can be repurposed for `cardano-db-sync` if the op
 
 ## Public endpoints
 
-Midnight-node uses partner-chains CLI commands which require on an Ogmios connect. While you may run Ogmios locally, we provide a public endpoint for the service.
-
-* [ogmios.testnet-02.midnight.network](https://ogmios.testnet-02.midnight.network)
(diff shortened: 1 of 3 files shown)


nstanford5:
This is a retired section of the docs and these pages are not served at the main site (indicated by the prefix `_`).

Closing this as won't do.

PR #1213, opened 2026-08-10
build(deps): consolidate five transitive dependency bumps

oduameh:
## What

Rolls five outstanding Dependabot transitive bumps into a single lockfile update:

| Package | From | To | Supersedes |
|---|---|---|---|
| websocket-driver | 0.7.4 | 0.7.5 | #1207 |
| svgo | 3.3.3 | 3.3.4 | #1208 |
| http-proxy-middleware | 2.0.9 | 2.0.10 | #1210 |
| @babel/core | 7.28.5 | 7.29.7 | #1211 |
| joi | 17.13.3 | 17.13.4 | #1212 |

## Why

Each of these bumps rebases and re-runs full CI on every merge to main, so merging them individually means one roughly eight minute CI cycle per package while Dependabot keeps opening more. One consolidated update lands the same lockfile state in a single cycle. Dependabot closes the superseded PRs automatically once this merges.

## Scope

`yarn.lock` only. No `package.json` changes, so no dependency ranges move. All five are transitive dependencies.

`serialize-javascript` and `shell-quote` had queued Dependabot jobs but were already at their resolved versions, so they produce no change here.

## Verification

`yarn install` and `yarn build` run clean locally against this lockfile. The build emits its usual pre-existing broken anchor warnings in the versioned `0.0.0` docs and the Compact reference, which are unrelated to dependency resolution and present on main.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -373,26 +373,33 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/compat-data@npm:^7.29.7":
+  version: 7.29.7
+  resolution: "@babel/compat-data@npm:7.29.7"
+  checksum: 10c0/47913f05e08a45a1c9df38c02b4b49e391005085b489432647a1abe112e5d9c75e3be8ea5972b7f6da4ec5d1339922ceb9ea02b8a25d4ed1cb8636e5261f344e
+  languageName: node
+  linkType: hard
+
 "@babel/core@npm:^7.21.3, @babel/core@npm:^7.25.9":
-  version: 7.28.5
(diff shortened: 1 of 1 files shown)


oduameh:
Needs one approving review to merge (self-approval is blocked by branch protection).

Once this lands, #1207, #1208, #1210, #1211 and #1212 become no-ops and Dependabot will close them automatically.

Verification so far: `build` green on CI and a clean local `yarn build`. The `run-LHCI-workflow` failure is a Lighthouse 429 rate-limit against the preview URL, the same flake that hit #1196 and is unrelated to these bumps.

PR #1218, opened 2026-08-10
docs: merge the funding guides into one verified assembly

oduameh:
## Intent

Second slice of the guides consolidation: merge the funding pages into one assembly, **Funding a wallet**, hosted at the existing `/guides/acquire-tokens` URL so its inbound links and redirects stay valid. One page deletion, one redirect, three inbound links repointed. Nothing else in the guides section is touched.

| Retired URL | Redirects to |
|---|---|
| `/guides/generating-dust-programmatically` | `/guides/acquire-tokens` (direct, no chain) |

## What changed

- `docs/guides/acquire-tokens.mdx` becomes the assembly: a faucet procedure, a Lace registration procedure, a wallet SDK registration procedure, and a merged troubleshooting reference. Net effect across the branch is roughly 400 insertions against 1,200 deletions: the SDK path keeps every step of the old tutorial's flow while dropping its duplicated 400-line "full script" block and its restated endpoint and token-model prose, which now link to `networks-and-environments`.
- `docs/guides/generating-dust-programmatically.mdx` deleted.
- Inbound links repointed in `networks-and-environments.mdx`, `configure-providers.mdx`, and `tokens/overview.mdx`.
- Terminology aligned with the protocol and SDK: you **register** NIGHT for DUST generation (`registerNightUtxosForDustGeneration`, `DustRegistration`). The old "Delegate tDUST" heading is gone, while Lace's own labels (**Generate tDUST**, designation) are kept where the UI is described.

## Why the content changed, not just moved

Every step of the SDK path was executed against Preprod before being written down, including a real faucet request and a full run on a fresh wallet. That corrected several things the old tutorial had wrong:

- **The faucet's confirmation message had changed.** The old page quoted "this should take about 2 mins. Its id is"; the live faucet emits "You will shortly receive 1000 tNight in your wallet. This is the transaction ID:". Updated to match.
- **The example's `package.json` was missing `@midnight-ntwrk/midnight-js-protocol`**, which the script imports. It resolved only through npm hoisting and would fail under pnpm or Yarn PnP. Added.
- **The claim that Node needs a WebSocket polyfill is false** on the guide's own supported floor: the full facade init connected using the native global WebSocket on Node 22+. The polyfill stays in the snippet for consistent behavior across versions, but is no longer presented as a requirement.
- **Faucet funds are visible before a wallet finishes syncing**, because a partial sync can only undercount. The arrival watcher therefore no longer gates on `isSynced`; only the registration step does, because that genuinely requires it.
- **Three different unmeasured timings for DUST generation** ("within minutes", "a minute or two") were carried over from the old page. Removed rather than reprinted; the guide now states only that generation begins once the registration lands.
- The old page's inline proof-server Compose file pinned image `8.0.3`, stale against the support matrix. The assembly links to the proof server guide rather than pinning a version, per the no-hardcoded-versions convention.
- The barrel package `@midnightntwrk/wallet-sdk@1.2.0` was confirmed by execution to export all twelve identifiers the guide imports, settling a scope-name ambiguity with `@midnight-ntwrk/wallet-sdk-facade` (both packages are real; the barrel's scope genuinely has no hyphen, which the guide now says out loud so it does not get "corrected").

## Verification

- **Executed test, pasted verbatim**: `wallet-construction.test.ts` (deterministic key derivation, Bech32m address prefix, and facade connectivity to all three Preprod wallets) passes 3/3 and is byte-identical to the block published in the guide.
- **Full journey run**: a fresh seed through faucet funding to a submitted registration transaction was executed end to end against Preprod; its findings shaped the procedure even though the log itself is not printed in the guide.
- **SME review**: approve with corrections; all three blockers (faucet text, polyfill claim, missing dependency) are fixed on this branch, along with two claims it flagged as unverifiable.
- **Style-guide review**: run as a separate pass after the SME review. It caught a screenshot alt text describing the wrong step, the inconsistent DUST timings, a missing Oxford comma, an admonition type, and several over-long sentences. Vale on the guide is now 0 errors, 0 warnings, and the Flesch grade sits under the house threshold.
- **Gates**: full site build green with the retired route absent and no broken links or anchors on any changed page; redirect audited as direct against all 300 rules; zero surviving references to the deleted page; `pipelines-urls.json` already lists the surviving slug.


oduameh:
Restructured the wallet SDK content after review feedback that the right-hand navigation felt thin compared with the page this replaces (855d0bcd).

The cause was structural rather than cosmetic. The old tutorial's ten `## Step N` headings gave it ten table-of-contents entries; folding all of it into one procedure turned those into numbered list items, which never appear in navigation. That same section was also covering three separate tasks, which the modular-docs standard says a procedure should not do. Both problems had one fix.

The wallet SDK content is now three procedures, each with the verification that actually belongs to it:

| Procedure | Verification |
|---|---|
| Setting up the wallet SDK project | proof server health check |
| Building a wallet from a seed | `wallet-construction.test.ts` (3 tests) |
| Registering NIGHT for DUST generation with the wallet SDK | `funding-helpers.test.ts` (3 tests) plus the end-to-end run output |

Table of contents goes from six entries to eight, and the three new ones were previously unreachable. Two anchors that the rename would have broken are repointed, including the inbound link from `configure-providers.mdx`.

The code still forms one script despite being presented in three parts: re-extracting the six procedure snippets across the new section boundaries and type-checking the concatenation still passes. Vale 0 errors, build green, no comment leakage in the rendered HTML.

Olanetsoft (approved):

PR #1221, opened 2026-08-11
docs: merge four deployment guides into deploy-and-operate

oduameh:
## Intent

Fourth slice of the guides consolidation: merge the four deployment-adjacent guides into one assembly that follows the reader's actual workflow, wire providers, deploy, observe, maintain. Four page deletions, one new page, and the section's last pre-consolidation redirect chain flattened.

| Retired URL | Redirects to |
|---|---|
| `/guides/configure-providers` | `/guides/deploy-and-operate` |
| `/guides/deploy-mn-app` | `/guides/deploy-and-operate` |
| `/guides/making-decision-on-contract-updatability` | `/guides/deploy-and-operate` |
| `/guides/query-the-blockchain` | `/guides/deploy-and-operate` |

Five older rules that pointed at these pages (`/guides/updatability`, `/guides/deployment`, `/getting-started/deploy-mn-app`, `/develop/tutorial/creating/updatability`, `/blog/query-the-blockchain`) are repointed at the new page directly, so every legacy URL resolves in a single hop; `/develop/tutorial/creating/updatability -> /guides/updatability -> page` was a pre-existing two-hop chain and is now flat. All 303 rules audited first-match, no shadowing, no duplicates.

Inbound links repointed in `security-best-practices`, `networks-and-environments`, and `install-bun-runtime-midnight`. One reference deliberately left alone: `generating-dust-programmatically.mdx` links `deploy-mn-app`, but that file is deleted by open PR #1218, so editing it here would manufacture a merge conflict; whichever PR lands second inherits a no-op.

## Why this shape

The four donors described four disconnected artifacts: a providers object, one example repo's Preprod script, an updatability API tour, and a Polkadot API walkthrough. The reader they share is one person doing one job: getting a tested contract onto a network and keeping it working. The assembly follows that job in order, five procedures with a concept module before each idea that needs one, and the reference tail is two lookup tables (provider implementations, troubleshooting) in the section's established genre.

## Verification: everything executed against a real network

The four verification blocks in the page are real Vitest files, byte-identical to the workspace copies, all executed against the local Midnight stack (node, indexer, proof server) started with the exact procedure the networks guide publishes: 4 files, 10 tests, all passing, twice. That includes a real `deployContract` (address returned, circuit called, state served back by the indexer) and a real maintenance cycle: `removeVerifierKey` after which the network rejects the circuit, `insertVerifierKey` after which it works again, and `replaceAuthority`.

Executing the donors' own code first surfaced four errors the published pages ship today:

1. **The `levelPrivateStateProvider` example throws as printed.** The current SDK requires an `accountId` option the donor never mentions: `accountId is required. Provide an account identifier (e.g., wallet address) to scope storage and prevent cross-account data access.`
2. **The donor's example password fails SDK validation.** `'your-encryption-password'` has two character classes; the SDK demands three of four: `PasswordValidationError: Password must contain at least 3 of: uppercase letters, lowercase letters, digits, special characters. Found: 2.`
3. **The donor's ledger-type imports fail to type-check.** Importing `ZswapSecretKeys` and friends from `@midnight-ntwrk/ledger-v8` directly collides with the copies midnight-js compiles against (`Types have separate declarations of a private property 'type_'`). The corrected pattern, imports from `@midnight-ntwrk/midnight-js-protocol/ledger`, passes strict tsc and matches the midnight-local-dev reference implementation.
4. **The donor never shows how to construct the `compiledContract` it passes to `deployContract`.** The missing bridge is `CompiledContract.make` + `withWitnesses` + `withCompiledFileAssets` from `@midnight-ntwrk/midnight-js-protocol/compact-js`, verified by the executed deploy.

Two exploration-era claims were softened by execution rather than repeated: the Polkadot `api.query.system.events.at(hash)` pattern still works on `@polkadot/api` 16.5.6 against a live node (the guide teaches the current `api.at(hash)` form anyway), and the query guide's real defects were eaten `\n` escapes and a "PRC" typo, not a dead API. The deploy donor's "delegate to start generating tDUST" is corrected to registration, matching the wallet SDK's `registerNightUtxosForDustGeneration` and the funding guide.

Every quoted message in the troubleshooting table was harvested by executing the failing call, including `Operation 'takeDown' is undefined for contract state` from a real verifier-key removal.

Not executed: the "Promoting a deployment to Preprod" procedure (a public-network deploy; its endpoints and funding flow are owned and verified by the networks and funding guides it links), and the `FetchZkConfigProvider` row (existence and export checked, not run).

## Notes for review

- The updatability concept distinguishes the ledger default (no authority, permanently non-upgradable) from the midnight-js default (`deployContract` samples and stores a single-signature authority when you pass no `signingKey`). The donor page only described the ledger default, which no longer matches what the SDK path gives you.
- PR #1126 edits `configure-providers.mdx`, which this PR deletes. Coordination comment posted there proposing its browser `levelPrivateStateProvider` procedure land as a module in this assembly once either merges.
- **Merge order: land #1218 (funding) and #1220 (testing) before this PR.** This page links `./acquire-tokens` as "Funding a wallet" (wallet SDK construction, DUST registration) and `./compact-javascript-runtime` as "Testing contracts off-chain"; on main those pages still have their pre-consolidation content, and only the open PRs carry the versions this page describes. No files are shared, so there is no conflict either way; only the `vercel.json` additions may need a trivial rebase.


docs/guides/deploy-and-operate.mdx
```diff
@@ -0,0 +1,804 @@
...
+ ✓ docs-tests/deploy.test.ts > deploying a contract > deploys and returns the contract address 20427ms
+ ✓ docs-tests/deploy.test.ts > deploying a contract > reconnects with findDeployedContract and calls a circuit 18675ms
+ ✓ docs-tests/deploy.test.ts > deploying a contract > reads the deployed state through the indexer 8ms
+
+ Test Files  1 passed (1)
+      Tests  3 passed (3)
+```
+
+{/* _mod-docs-content-type: PROCEDURE */}
+## Promoting a deployment to Preprod
```
Olanetsoft, line 467: This section doesn't indicate what we want to promote to preprod. I think we can hyperlink the hello world in the getting started section here.

docs/guides/deploy-and-operate.mdx
```diff
@@ -0,0 +1,804 @@
...
+2. Keep the wallet secret out of your source. Load the seed or mnemonic from an environment file that your version control ignores:
+
+   ```bash
+   # .env.preprod  (listed in .gitignore; never commit or share it)
+   MIDNIGHT_PREPROD_SEED=...   # 64 hex characters, or a 24-word mnemonic variable
+   ```
+
+3. Point the configuration at Preprod: `setNetworkId('preprod')`, the Preprod indexer and node endpoints from the [environment reference](./networks-and-environments#environment-reference), and a proof server you run yourself, following [Run a proof server](./run-proof-server). The proof server sees witness data in the clear, so it stays local even when the network is public.
+
+4. Run your deploy script. The first sync of a fresh Preprod wallet walks the chain's history and takes substantially longer than the local network's.
```
Olanetsoft: Especially this step. What script are we referring to? if its the hello world in the getting started section is reference then user can get the script from there. Do you understand what i mean?

Olanetsoft: Let me know what you think

docs/guides/deploy-and-operate.mdx
```diff
@@ -0,0 +1,804 @@
...
+
+### Verification
```
Olanetsoft, line 490: I noticed this pattern here and in the other PR. Do we really need this verification? I'm curious, as it doesn't do much for me as a user.

Since it's untestable, I'm not sure a user can just copy and paste it to their machine and have it compile without other setup already in place. That somehow defeats the purpose of a guide. I'm happy to be corrected here, but that's how I think about it.

docs/guides/deploy-and-operate.mdx
oduameh: Agreed, the procedure never named its object. It now opens by promoting the deploy from the preceding procedure explicitly, and links the hello world tutorial with a note that its repository packages this exact flow as `yarn test:preprod` for readers who started there.

docs/guides/deploy-and-operate.mdx
oduameh: Fair, "your deploy script" pointed at nothing. The step now says to run the same deploy code from the Deploying a contract procedure against the new configuration, and the intro covers the hello world path: if a reader started from that repo, `yarn test:preprod` is the packaged version of these steps.

docs/guides/deploy-and-operate.mdx
oduameh: The Verification sections are deliberate, and they are the part of these merged guides I would most want to keep, so let me make the case and also concede your point.

The case: every procedure ends with how you know it worked, and for the code ones the pasted files are not illustrative snippets, they were executed against a live local stack before publishing. That discipline is what caught the errors listed in the PR description: the published providers page we are replacing shipped an example that throws as printed (`accountId is required`), a password the SDK rejects, and imports that fail to type-check. Nothing catches that class of rot except running the code, and pasting the run is how the page proves it.

Your point is fair though: a reader could not tell from the page how to actually run them. That is fixed rather than removed: the first verification now states the run context and the exact command (`npx vitest run docs-tests ...` inside the midnight-local-dev clone the networks guide sets up), so copy-paste-run works verbatim. This particular verification you commented on is prose (an explorer lookup) precisely because a public-network deploy is a human check, not something docs can assert.

Olanetsoft (approved):

PR #1225, opened 2026-08-11
docs: fix two non-compiling snippets in keeping-data-private

Jc-asastu:
Two code examples in `docs/concepts/how-midnight-works/keeping-data-private.mdx` do not compile as written:

**1. `MerkleTreePath` has no `.value` field.**
`merkleTreePathRoot<#n, T>(path: MerkleTreePath<n, T>)` takes the path itself, and the struct only declares `leaf` and `path`. The second example in the same page already passes the path directly (`merkleTreePathRoot<10, Bytes<32>>(authPath)`), so this looks like a leftover.

**2. Pre-0.16 `assert` syntax.**
`assert !cond "msg";` predates the change to `assert(cond, "msg")`. The same code block already uses the current form a few lines above, so one of the two would fail to compile. This is the only occurrence left under `docs/`.

This page is the canonical explanation of the commitment/nullifier pattern, so the examples are likely to be copied as-is.

### DIFF
--- a/docs/concepts/how-midnight-works/keeping-data-private.mdx
+++ b/docs/concepts/how-midnight-works/keeping-data-private.mdx
@@ -147,7 +147,7 @@ export circuit insert(item: Field): [] {
 
 export circuit check(item: Field): [] {
   const path = findItem(item);
-  assert(items.checkRoot(merkleTreePathRoot<10, Field>(path.value)), "path must be valid");
+  assert(items.checkRoot(merkleTreePathRoot<10, Field>(path)), "path must be valid");
 }
 ```
 
@@ -209,7 +209,7 @@ export circuit increment(): [] {
   assert(authorizedCommitments.checkRoot(merkleTreePathRoot<10, Bytes<32>>(authPath)),
     "not authorized");
   const nul = nullifier(sk);
-  assert !authorizedNullifiers.member(nul) "already incremented";
+  assert(!authorizedNullifiers.member(nul), "already incremented");
   authorizedNullifiers.insert(disclose(nul));
   restrictedCounter.increment(1);
 }


Jc-asastu:
Note: this contribution was AI-assisted. Every change was verified against the source before submitting. Happy to have the `ai-assisted` label applied if that's your convention — I can't set labels as an outside contributor.

docs/concepts/how-midnight-works/keeping-data-private.mdx
```diff
@@ -147,7 +147,7 @@ export circuit insert(item: Field): [] {
 
 export circuit check(item: Field): [] {
   const path = findItem(item);
-  assert(items.checkRoot(merkleTreePathRoot<10, Field>(path.value)), "path must be valid");
+  assert(items.checkRoot(merkleTreePathRoot<10, Field>(path)), "path must be valid");
```
oduameh, line 150: Technical: the `.value` fix is right, but this line still fails to compile. `path` comes from the `findItem` witness, and a witness-derived value entering a ledger operation needs `disclose()`.

```suggestion
  assert(items.checkRoot(disclose(merkleTreePathRoot<10, Field>(path))), "path must be valid");
```

docs/concepts/how-midnight-works/keeping-data-private.mdx
```diff
@@ -209,7 +209,7 @@ export circuit increment(): [] {
   assert(authorizedCommitments.checkRoot(merkleTreePathRoot<10, Bytes<32>>(authPath)),
     "not authorized");
```
oduameh, line 210: Technical: same disclosure error here, via the `findAuthPath` witness. Rewrapped to keep the line readable.

```suggestion
  assert(authorizedCommitments.checkRoot(
    disclose(merkleTreePathRoot<10, Bytes<32>>(authPath))), "not authorized");
```

docs/concepts/how-midnight-works/keeping-data-private.mdx
```diff
@@ -209,7 +209,7 @@ export circuit increment(): [] {
...
   const nul = nullifier(sk);
-  assert !authorizedNullifiers.member(nul) "already incremented";
+  assert(!authorizedNullifiers.member(nul), "already incremented");
```
oduameh, line 212: Technical: the assert syntax is correct, but this line needs `disclose()` on the read to compile. `member(nul)` takes a value derived from the `secretKey` witness into a ledger operation. Line 213 already does this for the write.

```suggestion
  assert(!authorizedNullifiers.member(disclose(nul)), "already incremented");
```

oduameh (changes requested):
Both claims verified by compilation: compact CLI 0.5.1, compactc 0.31.1, language version 0.23.0.

Claim 1 confirmed. The compiler prints `MerkleTreePath` as two fields, `leaf` and `path`, and rejects the current code with `structure MerkleTreePath has no field named value`. Matches `docs/compact/standard-library/exports.md:338`, `circuit merkleTreePathRoot<#n, T>(path: MerkleTreePath<n, T>): MerkleTreeDigest`.

Claim 2 confirmed: `parse error: found "!" looking for "("`. "Pre-0.16" is accurate. Language version 0.15.0 was the last to accept `assert cond "msg"`; 0.16.0 introduced the call form, and no version accepts both. The other hit under `docs/`, `relnotes/compact/compact-0-16-24-0.mdx:122`, is the intentional "before" example in the release note announcing that change and should stay as it is.

Blocker: after these fixes neither snippet compiles. Both defects were masking disclosure errors, because compilation aborts before disclosure analysis runs. Clearing them surfaces five more, two of them in circuits this PR does not touch.

Snippet 1:

```compact
export circuit insert(item: Field): [] {
  items.insert(disclose(item));
}

export circuit check(item: Field): [] {
  const path = findItem(item);
  assert(items.checkRoot(disclose(merkleTreePathRoot<10, Field>(path))), "path must be valid");
}
```

Snippet 2:

```compact
export circuit addAuthority(pk: Bytes<32>): [] {
  authorizedCommitments.insert(disclose(pk));
}

export circuit increment(): [] {
  const sk = secretKey();
  const authPath = findAuthPath(publicKey(sk));
  assert(authorizedCommitments.checkRoot(
    disclose(merkleTreePathRoot<10, Bytes<32>>(authPath))), "not authorized");
  const nul = nullifier(sk);
  assert(!authorizedNullifiers.member(disclose(nul)), "already incremented");
  authorizedNullifiers.insert(disclose(nul));
  restrictedCounter.increment(1);
}
```

Three of the five are inside the diff and have suggestions. Two are outside it and need editing by hand:

- Line 145: `items.insert(item);` becomes `items.insert(disclose(item));`
- Line 203: `authorizedCommitments.insert(pk);` becomes `authorizedCommitments.insert(disclose(pk));`

Line 212 is the line this PR already edits. `member(nul)` reads a witness-derived value into a ledger operation, so the read needs `disclose()`, not only the `insert` on line 213.

Parameters of an exported circuit count as witness values, which is why `insert(item)` and `insert(pk)` fail.

Verified with all five applied: both contracts compile clean and execute correctly. A valid path passes, a tampered path fires `failed assert: path must be valid`, and a replayed `increment()` fires `failed assert: already incremented` with the counter holding at 1.

Non-blocking:

- Neither block declares `pragma language_version`. It compiles without one, but this is the canonical page for the commitment/nullifier pattern.
- `merkleTreePathRoot` returns `MerkleTreeDigest`, not `Field`, if the surrounding prose is ever revised.

wbaxterh:
Rebase-ready fix for both snippets is small, and oduameh's three disclose() suggestions above are exactly what the current compiler wants; I hit the same three errors porting this pattern in my own contract. @Jc-asastu if you are busy, I am happy to carry this across the line in a follow-up PR with you credited as co-author. I will wait a few days before doing anything.

notsoblack:
## Review: Compact snippet fixes in keeping-data-private\n\nVerified both changes against the Compact language spec:\n\n1. **`path.value` → `path`** ✅ Correct. `findItem(item)` returns `MerkleTreePath<...>` directly; `merkleTreePathRoot<10, Field>()` expects the path object, not `.value`.\n2. **`assert !member(...) "text"` → `assert(!member(...), "text")`** ✅ Correct. Compact `assert` requires parentheses and comma-separated arguments. The original would fail to compile.\n\n**Scope check:** Only the two intended lines changed. No collateral.\n\n**Sign-off:** LGTM — legitimate community contribution fixing real non-compiling examples.

wbaxterh:
Opened #1314, which carries this forward with you as `Co-authored-by` on the commit.

**Your two fixes are in it unchanged**, along with @oduameh's three `disclose()` suggestions applied as written.

**It also fixes two more lines the review did not reach:** `items.insert(item)` and `authorizedCommitments.insert(pk)`. Both are exported-circuit parameters flowing into a ledger write, so the compiler wants `disclose()` even though they look public. I pulled both `compact` blocks out of the page verbatim and compiled them: with this PR and the three review suggestions applied, the page still fails on those two.

**Plus a binding assert in each example**, `path.leaf` against the value the circuit computed. Without it `checkRoot` only proves some leaf is in the tree, not yours, and in `increment` an observed path can be replayed under an unauthorized key.

One thing worth knowing if you verify locally: `compact compile` prints `Exception:` on failure but still exits 0, so check the output or the emitted keys rather than the exit code.

Happy to close #1314 in favour of this one if you would rather finish it yourself. Just say the word.

notsoblack:
Perfect, thank you . You can close it now.

PR #1244, opened 2026-08-18
docs(tutorials): add NFT-ZK Undeployed on Fly.io

arunnadarasa:
## Summary

Adds an intermediate tutorial that covers two gaps in the current docs:

- Hosting the Undeployed node, indexer, and proof server on Fly.io. Existing guides only cover Docker Desktop on localhost.
- Deploying and minting the official NFT-ZK contract with real admin-keypair authorization. The example-nft-contracts simulator tests never put the contract on a real Undeployed node.

The proof-server page documents the IPv4-only binary versus Fly 6PN (IPv6-only) constraint that breaks prove requests to `*.internal`. The mint page documents constructor admin-keypair checks (not an empty-hash assert), insert-only Compact maps, the reserved `list` keyword, and RpcError 117 private-state mismatch.

Image tags match [midnight-local-dev `standalone.yml`](https://github.com/midnightntwrk/midnight-local-dev/blob/main/standalone.yml) as of August 2026 (`midnight-node:1.0.0`, `indexer-standalone:4.3.3`, `proof-server:8.1.0`). Fly configs are inlined so the tutorial does not depend on unpublished scripts.

## Test plan

- [x] `vale docs/tutorials/nft-zk-fly/` (0 errors, 0 warnings; remaining suggestions are the Apache license header passive voice)
- [x] Local Docusaurus preview: `/tutorials`, `/tutorials/nft-zk-fly`, `/tutorials/nft-zk-fly/host-on-fly`, `/tutorials/nft-zk-fly/mint-nft-zk` all return 200
- [ ] Maintainer preview after Vercel authorization
- [ ] Confirm sidebar category `NFT-ZK on Fly` and Tutorials landing tile render

Made with [Cursor](https://cursor.com)


nstanford5:
Thanks for the submission @arunnadarasa -- a few things:

1. We explicitly do not recommend hosting the proof server anywhere -- it should be run on users local hardware only. We will not be publishing a tutorial that supports this as it goes against infra best practice for developing mn apps

2. `_intermediate` is a hidden area of the docs, not served to the site. Historical, can probably be removed @oduameh 

3. Closing this as won't do

PR #1252, opened 2026-08-20
docs: fix dead testnet-02 RPC endpoint in run-a-validator guide

### LINKED ISSUE #1172: ~80 stale testnet-02 references across org repos, six months after the Preview transition

CjDabrow:
Fixes #1172.

Four example commands in the run-a-validator guide pointed at `rpc.testnet-02.midnight.network`, which was retired with testnet-02 and no longer resolves (NXDOMAIN). This PR updates them to the current live `rpc.preview.midnight.network` endpoint, matching the Preview onboarding guide.

Also, per review feedback: the first `sidechain_getStatus` example passed `"params": [INSERT_EPOCH]`, but the method accepts no parameters — changed to `"params": []` (the `systemParameters_getAriadneParameters` examples keep their epoch parameter, which that method does take).

**Verification:**
- `host rpc.testnet-02.midnight.network` → NXDOMAIN
- `host rpc.preview.midnight.network` → resolves (3 A records)
- Docs-only change, single file: `docs/nodes/_run-a-validator/step-3.mdx`

*Apologies for the earlier PR description — it was pasted from an unrelated project by mistake and described work that is not in this repo. This body now describes the actual diff.*

### DIFF
--- a/docs/nodes/_run-a-validator/step-3.mdx
+++ b/docs/nodes/_run-a-validator/step-3.mdx
@@ -428,7 +428,7 @@ curl -L -X POST -H "Content-Type: application/json" -d '{
       "method": "sidechain_getStatus",
       "params": [INSERT_EPOCH],
       "id": 1
-    }' https://rpc.testnet-02.midnight.network | jq
+    }' https://rpc.preview.midnight.network | jq
 ```
 
 Example output:
@@ -439,7 +439,7 @@ curl -L -X POST -H "Content-Type: application/json" -d '{
       "method": "sidechain_getStatus",
       "params": [],
       "id": 1
-    }' https://rpc.testnet-02.midnight.network | jq
+    }' https://rpc.preview.midnight.network | jq
   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                  Dload  Upload   Total   Spent    Left  Speed
 100   297  100   193  100   104    395    213 --:--:-- --:--:-- --:--:--   608
@@ -469,7 +469,7 @@ curl -L -X POST -H "Content-Type: application/json" -d '{
       "method": "systemParameters_getAriadneParameters",
       "params": [697],
       "id": 1
-    }' https://rpc.testnet-02.midnight.network | jq
+    }' https://rpc.preview.midnight.network | jq
 ```
 
 Look for your registration under `"candidateRegistrations": {)`. Ensure you see your Partner-chain keys in the registration and  `"isValid": true`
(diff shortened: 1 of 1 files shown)


docs/nodes/_run-a-validator/step-3.mdx
```diff
@@ -428,7 +428,7 @@ curl -L -X POST -H "Content-Type: application/json" -d '{
       "method": "sidechain_getStatus",
       "params": [INSERT_EPOCH],
```
oduameh: Technical: `sidechain_getStatus` takes no parameters, and the example output below already uses `[]`. The endpoint tolerates the extra argument, so this is cosmetic.

```suggestion
      "params": [],
```

oduameh (changes requested):
**The PR description does not match the change.** The body describes a NightSign receipt provenance fix (`/api/verify`, `receipt_origin`, `src/types/index.ts`), none of which is in this repo. The diff swaps `rpc.testnet-02.midnight.network` for `rpc.preview.midnight.network` in four curl commands, which is correct. Replace the description with one that describes this change.

oduameh:
cc @stevanlohja

CjDabrow:
@oduameh Both points addressed: (1) the PR description was pasted from an unrelated project by mistake — it now describes the actual diff; (2) the `sidechain_getStatus` example now uses `"params": []` (pushed in f9c214e; the `systemParameters_getAriadneParameters` examples keep their epoch parameter, which that method does take). Ready for another look — thanks for the catch.

oduameh:
Thanks for the quick turnaround, the new description reads exactly right and the `params` fix is in. One last thing before this can merge: the repo requires verified commit signatures, and both commits on the branch are currently unsigned. The simplest route is to squash them into one commit with signing enabled (`git rebase` onto main, then `git commit -S`, or set `git config commit.gpgsign true` and amend) and force-push the branch. Happy to approve as soon as that's done.

oduameh:
Quick heads-up: #1304 now touches the same step-3 file (plus other pages), and I have asked its author to rebase around your endpoint fixes since yours came first and is already approved on content. So this is still just waiting on the signed commit from you, and landing it soon keeps the rebase simple for everyone. Shout if the signing steps are giving you any trouble, happy to help.

PR #1257, opened 2026-08-21
docs: add browser levelPrivateStateProvider guide and canary password verification

pplmaverick:
## Summary

Adds two pieces of browser-specific guidance to `docs/guides/deploy-and-operate.mdx`, re-targeting content from PR #1126 (which was opened against the now-deleted `configure-providers.mdx`).

## Changes

**1. Browser environments note** (after Step 3 of "Configuring providers for a contract")

Explains that `cryptoBackend` should be left unset so the SDK's Noble fallback stays available in non-secure contexts and React Native. Clarifies that `level` resolves to `browser-level` (IndexedDB) only in client bundles, not in Next.js server components or Edge functions.

**2. Canary password verification** (new `###` section under "Provider implementations")

Provides a `provisionCanary` / `verifyPassword` split that avoids the bootstrapping flaw where a missing canary and a wrong password are indistinguishable at read time. Includes the `PrivateStateProvider` import, preserves the underlying error via `cause`, and distinguishes a missing canary from a bad password.

Both changes verified against `@midnight-ntwrk/midnight-js-level-private-state-provider@4.1.1` and `5.0.0-beta.6`.

Closes #1126.


docs/guides/deploy-and-operate.mdx
```diff
@@ -99,6 +99,23 @@ Construct the six providers and assemble them into the object every deploy and c
...
+   `levelPrivateStateProvider` works in browser environments. The SDK resolves the crypto
+   backend automatically: it uses Web Crypto where `crypto.subtle` is available and falls back to
+   a pure-JS implementation where it is not. Leaving `cryptoBackend` unset is the safer default,
+   because requesting `'webcrypto'` explicitly turns an unavailable Web Crypto API into a thrown
+   error rather than a fallback. That matters in non-secure contexts, such as a page served over
+   plain `http://` on a non-loopback host, and in React Native.
+
+   The `level` package resolves to `browser-level` (IndexedDB) in browser bundles through its
+   package.json `browser` field, which Vite and webpack honor automatically, so no extra storage
+   configuration is needed. Keep private state access on the client: code that runs outside a
```
oduameh: Our style linter prefers the active voice here, tiny tweak:

```suggestion
   package.json `browser` field, which Vite and webpack honor automatically, so you do not need
   extra storage configuration. Keep private state access on the client: code that runs outside a
```

docs/guides/deploy-and-operate.mdx
```diff
@@ -781,6 +798,51 @@ Which implementation fills each provider slot, by environment. Packages live und
...
+
+Without an explicit check, a wrong password is not caught until the app attempts to decrypt
+real private state data, which may happen deep inside a transaction flow. The failure also
+looks different depending on the active backend: Web Crypto rejects with a `DOMException`
+named `OperationError`, while the pure-JS backend throws a plain `Error`.
+
+A canary value written at account creation lets you catch a wrong password at unlock time
+with a clear message. Split provisioning from verification: a missing canary and a wrong
+password are indistinguishable at read time, so bootstrapping the canary from a `null` read
+would certify the failure it was added to catch.
```
oduameh: Same here:

```suggestion
would certify the failure it exists to catch.
```

oduameh (changes requested):
This looks great. I checked both sections against the SDK source on 4.1.1 and the 5.0 beta, and everything you wrote holds up, including the subtle bits: the explicit `'webcrypto'` opt-in really does throw where the default quietly falls back, and `get()` gives you `null` for a missing key but rejects on a bad password. The snippet compiles clean under strict TypeScript too, even with a fully typed provider. Nice work on the provisioning split, that was the hard part of this one.

A few housekeeping things before merge:

1. Your commit isn't signed, and the repo needs verified signatures. Amending with signing on and force-pushing will do it.
2. The branch is a bit behind main, worth a quick update.
3. Small heads-up: "Closes #1126" only auto-closes issues, not PRs, so the old one will stay open. I'll close it myself once this lands, nothing for you to do there.

I left two tiny wording suggestions inline. Once the signature is sorted I'm happy to approve.

pplmaverick:
Both wording suggestions applied and commits re-signed. Ready for re-review.

oduameh (approved):
Both wording tweaks are in and the commits are signed and verified, so this is good to go from my side. Thanks for a genuinely strong guide, the canary provisioning split especially. I'll close #1126 once this lands, as promised.

oduameh (approved):

PR #1275, opened 2026-08-27
Add compact-js v2.5.3 release notes

MidnightCI:
## Summary

Adds the Compact.js 2.5.3 release notes (the docs stopped at 2.5.0) and fills in the DynamicList entry.

While preparing this I found the 2.5.1/2.5.2 lineage is messy: 2.5.1 went to npm with no GitHub release, 2.5.2 was tagged in the migrated `midnightntwrk/midnight-sdk` repo but never published to npm, and the release changelog sections are cumulative copies of each other. So the page documents 2.5.3 as the maintenance release that supersedes both (CommonJS exports fix, `tar` security bump, dependency and release automation updates) and says so explicitly, rather than inventing separate 2.5.1/2.5.2 pages.

## Notes for review

- The support matrix references `compact-js-2.5.1`, which exists on npm but not as a git tag. If the matrix should point at 2.5.3, that is a follow-up with SME confirmation.


oduameh:
cc @nstanford5

nstanford5 (approved):

PR #1276, opened 2026-08-27
Add dapp-connector-api v4.0.1 release notes

MidnightCI:
## Summary

Adds the DApp Connector API v4.0.1 release notes, undocumented since February even though the support matrix already lists 4.0.1. Single change: the `payFees` option, previously only on `makeIntent`, is now available on all transacting methods (backwards-compatible), which enables sponsored-fee flows where a separate wallet pays fees.

No matrix change needed: it already shows 4.0.1.


oduameh:
cc @nstanford5

nstanford5 (approved):

PR #1278, opened 2026-08-27
docs(guides): add Vite configuration section to React wallet connector guide

Emmanuellsensai:
## What and why

The React wallet connector guide tells developers to scaffold with Vite
but does not show the required `vite.config.ts` setup for Midnight's
WASM packages. Developers following the guide hit a blank page in
development with no visible console error.

## What this PR adds

A "Configure Vite for Midnight packages" section between the install
step and the tutorial steps, covering:

- `optimizeDeps.exclude` for the three WASM packages
- Why pre-bundling breaks wasm-bindgen's top-level init order
- A note that `vite-plugin-wasm` must NOT be added on Vite 8
- A pointer to the resolve.dedupe guide for monorepo setups

The configuration is validated from a real Midnight dApp
(Anonymous Whispers) built on the same Vite + React stack.

## Related
- midnight-js#1052 (WASM dual-instantiation, docs merged via PR#1129)


Emmanuellsensai:
HI @oduameh @nstanford5  Happy to make any adjustments needed. This came from a real blank-page issue I hit building a Midnight dApp with Vite + React

docs/guides/react-wallet-connect.mdx
```diff
@@ -40,6 +40,36 @@ Then install the DApp Connector API package:
 ```bash
 npm install @midnight-ntwrk/dapp-connector-api
 ```
+### Configure Vite for Midnight packages
```
oduameh, line 44: Blank line between the code fence and the heading, so markdown parsers do not glue them together:

```suggestion

### Configure Vite for Midnight packages
```

docs/guides/react-wallet-connect.mdx
```diff
@@ -40,6 +40,36 @@ Then install the DApp Connector API package:
...
+
+Midnight SDK packages ship WASM modules (via wasm-bindgen) that require specific Vite configuration. Without this step, your app will render a blank page in development or fail with `Cannot access '__wbindgen_start' before initialization`.
```
oduameh: The wallet connector built on this page never loads these packages, so as written this promises a failure the reader will not hit. Reframing it as preparation for the SDK step keeps the advice and makes it accurate:

```suggestion
The wallet connector in this guide uses only the DApp Connector API, which ships no WASM. As soon as your app pulls in Midnight's SDK packages to work with contracts, such as Midnight.js and its ledger bindings, their wasm-bindgen modules need specific Vite configuration. Without it, the app renders a blank page in development or fails with `Cannot access '__wbindgen_start' before initialization`. Adding the configuration now keeps your project ready for that step.
```

oduameh (changes requested):
Thanks for this, the underlying advice is real and well researched. I verified the pieces: all three excluded package names exist on npm, the linked vite-wasm-resolution guide in midnight-js is live, and the dedupe note for monorepos matches what that guide says. The caution box explanation of the optimizer reordering `__wbindgen_start()` is a genuinely useful bit of context that I have not seen written down in our docs anywhere.

One accuracy problem before this can merge, though. The app this guide builds never hits the issue the section warns about. The only package it installs is `@midnight-ntwrk/dapp-connector-api`, which is a types-only package with zero dependencies and no WASM; the connector talks to the wallet extension through `window.midnight`. So "without this step, your app will render a blank page" is not true for a reader who follows just this guide, and they may wonder why excluding three packages they never installed does anything. The fix is small: frame the section as preparing the project for the moment they add the SDK packages that do ship WASM. I have left a suggestion on the intro paragraph doing exactly that, plus a tiny one adding the blank line before the heading so the fence and heading do not collide.

Two smaller things:

1. Placement: the section currently lands between the install command and the "After completing this tutorial, you'll understand" list, which leaves that intro list dangling after the caution box. Moving your section to just after that list would keep the guide's opening intact.
2. The PR description mentions a note about not adding `vite-plugin-wasm` on Vite 8, but the diff does not contain it. Either version works for me, they just want to agree with each other.

Happy to re-review quickly once these are in.

Emmanuellsensai:
Hi @oduameh , all three items are addressed:

1. Blank line added before the heading so the code fence and heading do not collide
2. Intro paragraph reframed -- the section now explains this is preparation for when SDK packages with WASM are added, not a requirement for this guide alone
3. A note about not adding vite-plugin-wasm on Vite 8 is now in the caution box, so the diff and PR description agree

Ready for re-review.

docs/guides/react-wallet-connect.mdx
```diff
@@ -41,11 +41,36 @@ Then install the DApp Connector API package:
...
+      '@midnight-ntwrk/ledger-v8',
+      '@midnight-ntwrk/onchain-runtime-v3',
+      '@midnight-ntwrk/zkir-v2',
+    ],
+  },
+});
+```
+
+:::caution Why this matters
+Vite's dev-server dependency optimizer eagerly bundles dependencies into single modules. Midnight's WASM glue code calls `__wbindgen_start()` at the module top level during initialization. Pre-bundling reorders this init sequence, producing a blank page with no visible error in the console. The `optimizeDeps.exclude` setting tells Vite to serve these packages unbundled, preserving the correct initialization order. Production builds are unaffected. Do not add `vite-plugin-wasm`; on Vite 8 it conflicts with this approach and is not required.
```
oduameh: One accuracy snag: the Vite WASM resolution guide this section links to uses `vite-plugin-wasm` in its own recommended config, so a blanket "do not add" here contradicts the reference one paragraph later. The Vite 8 incompatibility that guide records is about a different plugin, `vite-plugin-top-level-await`. If your Anonymous Whispers build genuinely hit a `vite-plugin-wasm` conflict on Vite 8, that is worth raising on the midnight-js guide so the two documents agree; until then, the safe wording is the neutral one.

```suggestion
Vite's dev-server dependency optimizer eagerly bundles dependencies into single modules. Midnight's WASM glue code calls `__wbindgen_start()` at the module top level during initialization. Pre-bundling reorders this init sequence, producing a blank page with no visible error in the console. The `optimizeDeps.exclude` setting tells Vite to serve these packages unbundled, preserving the correct initialization order. Production builds are unaffected. This setup needs no extra WASM plugins.
```

docs/guides/react-wallet-connect.mdx
```diff
@@ -41,11 +41,36 @@ Then install the DApp Connector API package:
...
+:::
+
+If you are working in a monorepo where multiple packages depend on `@midnight-ntwrk/compact-runtime`, also add `resolve.dedupe` to prevent WASM dual-instantiation (see the [Vite WASM resolution guide](https://github.com/midnightntwrk/midnight-js/blob/main/docs/guides/vite-wasm-resolution.md) for details).
```
oduameh, line 73: The placement fix accidentally removed the "After completing this tutorial" outcomes list rather than moving it. Restoring it here keeps the guide's opening promise intact.

```suggestion
If you are working in a monorepo where multiple packages depend on `@midnight-ntwrk/compact-runtime`, also add `resolve.dedupe` to prevent WASM dual-instantiation (see the [Vite WASM resolution guide](https://github.com/midnightntwrk/midnight-js/blob/main/docs/guides/vite-wasm-resolution.md) for details).

After completing this tutorial, you'll understand:

- The Midnight wallet connection flow
- How to use the DApp Connector API
- Best practices for managing wallet state in React applications
```

oduameh (changes requested):
Thanks for the quick turnaround, the reframed intro reads exactly right and the blank-line fix is in. Two things surfaced in this round, both small:

1. The placement change deleted the "After completing this tutorial" outcomes list instead of keeping it; there is a suggestion restoring it.
2. The new `vite-plugin-wasm` sentence contradicts the guide it links to, which itself uses that plugin; the Vite 8 problem documented there belongs to `vite-plugin-top-level-await`. Suggestion attached with the details.

One housekeeping item for the end: the repo needs verified commit signatures, and your two locally authored commits are currently unsigned (the suggestion commit applied through GitHub is fine). Squashing the branch into one signed commit once the content settles will sort it.

This is genuinely close now, happy to approve as soon as these land.

Emmanuellsensai:
i have applied the suggested changes @oduameh 
ready for review

PR #1289, opened 2026-08-31
Add Wallet SDK 1.0.0, 1.1.0, and 1.2.0 release notes (umbrella version line)

oduameh:
## Summary

Backfills the three undocumented stable Wallet SDK releases, now on the **umbrella version line** per the review discussion: **Wallet SDK 1.0.0** (23 April), **1.1.0** (26 May), and **1.2.0** (1 July, first release under the `@midnightntwrk` scope). Content is drawn from the sub-package changelogs, since the umbrella's own release notes are republish markers.

Also in this PR, so the scheme change lands atomically:

- `DynamicListWallet.js` now defaults to the curated top entry instead of the numeric maximum, and dedupes the version dropdown. Without this, the legacy facade-numbered entries (1.0.0 to 3.0.0) would win the default view over the new umbrella 1.2.0.
- The wallet release notes landing page gets a short note marking the numbering scheme change.
- `scripts/check-release-gaps.mjs` (the release watcher) now tracks the `@midnight-?ntwrk/wallet-sdk@x.y.z` umbrella tags for its Wallet SDK row; dry run confirms it reports current against this branch. Gap issue #1300 ("Wallet SDK 4.1.0", filed under the old facade tracking) can close when this merges.

Closes #1300
Closes #1237

## Follow-up (separate PR, as agreed)

Point the wallet dev guide's install and imports at the umbrella package, matching what the example code already does.

PR #1274, opened 2026-08-27
Add ledger v8.1.1 release notes

MidnightCI:
## Summary

Adds the ledger 8.1.1 release notes (stable since 4 August) and fills in the DynamicList entry. Patch release: array-handling test coverage, clippy cleanups, no protocol or API behavior changes.

The one substantive item is the npm scope change: ledger packages now publish under `@midnightntwrk` (previously `@midnight-ntwrk`). Both scopes currently resolve to 8.1.1 on npm.

## Notes for review

- The support matrix still shows ledger 8.1.0 for all networks. Whether 8.1.1 is the deployed/tested version is an SME call, so the matrix is untouched here.
- The scope change has wider docs implications: the style guide and most pages standardize on `@midnight-ntwrk/`. If other components follow the rename, we need a coordinated docs sweep. Flagging rather than acting on it in this PR.

PR #1290, opened 2026-08-31
ci: watch product repos for undocumented stable releases

oduameh:
## Summary

Closes the detection hole in the release notes process: today nothing tells us when a product repo ships, so gaps are found by manually diffing (which is how the recent node/ledger/wallet backfills were discovered, and how the wallet one still slipped a first audit).

Adds a daily `release-watch.yml` workflow backed by `scripts/check-release-gaps.mjs`:

- For each of nine components (node, ledger, indexer, Midnight.js, DApp Connector, Compact toolchain, Compact devtools, Compact JS, Wallet SDK facade), it fetches the product repo's releases and keeps only stable ones. Two filters are applied deliberately: the `prerelease` flag AND a clean `x.y.z` tag pattern, because midnight-js has betas not flagged as prereleases and the indexer flags clean-tagged builds as prereleases. Both real-world quirks are covered.
- The highest stable version is compared against the newest entry in the component's DynamicList.
- Each gap opens one issue titled `Release notes gap: <component> <version>` with the release link and the resolution steps. An existing issue with that title, open or closed, is never recreated, so closing an issue as "won't document" silences that version permanently.
- A coverage table lands in the workflow step summary on every run, and `workflow_dispatch` has a dry-run input for testing.

## Validation

Dry run against current main detects exactly the two known gaps and nothing else:

| Component | Documented | Upstream stable | Status |
|---|---|---|---|
| Ledger | 8.1.0 | 8.1.1 | gap (PR #1274 in review) |
| Wallet SDK | 3.0.0 | 4.1.0 | gap (PR #1289 in review) |
| Other seven | current | current | ok |

## Scope notes

- Proof server and on-chain runtime have no public release tags; Lace is a private repo the workflow token cannot read; pubsub-indexer and examples are dormant; faucet's release hygiene is too noisy to watch usefully (an alpha is currently tagged Latest). All skipped deliberately.
- Permissions are minimal (`contents: read`, `issues: write`), actions are SHA-pinned per repo convention, and the wallet facade tag regex matches both the old `@midnight-ntwrk` and new `@midnightntwrk` scopes.

### DIFF
--- a/.github/workflows/release-watch.yml
+++ b/.github/workflows/release-watch.yml
@@ -1,0 +1,40 @@
+# This file is part of midnight-docs.
+# Copyright (C) Midnight Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+# Watches the Midnight product repos for stable releases that have no release
+# notes in the docs, and opens one issue per missing component version.
+# Detection logic lives in scripts/check-release-gaps.mjs.
+
+name: Release notes watch
+
+on:
+  schedule:
+    - cron: '17 7 * * *' # daily, 07:17 UTC
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Report gaps without opening issues'
+        required: false
+        default: false
+        type: boolean
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  check-release-gaps:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


oduameh:
Heads-up for review: the Wallet SDK row currently watches the facade tags. Depending on where the versioning discussion on #1289 lands (facade line vs the wallet-sdk umbrella line), the wallet regex here will follow suit; one-line change either way.

nstanford5 (approved):
Solid workflow improvement closing a real gap. Great work @oduameh

PR #1304, opened 2026-09-03
docs: replace stale testnet-02 endpoints with Preview in validator + wallet guides

notsoblack:
Partial fix for #1172 — covers the stale `testnet-02` references that live in this repo (`midnight-node-docker`'s README/.envrc are separate).

`rpc.testnet-02.midnight.network` / `indexer.testnet-02.midnight.network` no longer resolve (NXDOMAIN since the 2026-01-27 Preview transition), which sends new integrators to a dead endpoint. This updates the user-facing guides.

**Changes**
- `nodes/run-a-validator/step-3.mdx` — point the validator-status `curl` queries at `rpc.preview.midnight.network`. Also fixes the `sidechain_getStatus` "Example output" block, which mistakenly contained the `curl` command rather than only the JSON response.
- `nodes/run-a-validator/step-1.mdx` — update the supported-environments table (`testnet-02` is retired; the flow targets the Preview-era testnet) and link to the [Networks and environments](/guides/networks-and-environments) reference.
- `nodes/run-a-validator/index.mdx` — remove the `ogmios.testnet-02.midnight.network` public-endpoint link (that domain no longer resolves); direct readers to the current endpoints page.
- `main_versioned_docs/version-0.0.0/develop/guides/wallet-dev-guide.mdx` — update the `WalletBuilder` code examples from the dead `testnet-02` endpoints (`api/v1`) to the current Preview endpoints (`api/v4`).

Release notes and other historical records that mention `testnet-02` are intentionally left as-is.

### DIFF
--- a/docs/nodes/_run-a-validator/index.mdx
+++ b/docs/nodes/_run-a-validator/index.mdx
@@ -171,7 +171,7 @@ An existing Cardano relay node can be repurposed for `cardano-db-sync` if the op
 
 Midnight-node uses partner-chains CLI commands which require on an Ogmios connect. While you may run Ogmios locally, we provide a public endpoint for the service.
 
-* [ogmios.testnet-02.midnight.network](https://ogmios.testnet-02.midnight.network)
+* The former public Ogmios endpoint (`ogmios.testnet-02.midnight.network`) has been retired along with the `testnet-02` network and no longer resolves. Run Ogmios locally, and see [Networks and environments](/guides/networks-and-environments) for the current endpoints.
 
 ### Network and Security Considerations for Testnet:
 
--- a/docs/nodes/_run-a-validator/step-1.mdx
+++ b/docs/nodes/_run-a-validator/step-1.mdx
@@ -13,7 +13,7 @@ To embark on the path of becoming a Midnight validator, one must either be, or b
 
 | Cardano env. | Status | Midnight env.
 |----------|----------|----------|
-| `preview` | ✅ | `testnet-02` |
+| `preview` | ✅ | `preview` (the `testnet-02` network has been retired; see [Networks and environments](/guides/networks-and-environments)) |
 | `preprod` | ❌ | N/A |
 | `mainnet` | ❌ | N/A |
 
--- a/docs/nodes/_run-a-validator/step-3.mdx
+++ b/docs/nodes/_run-a-validator/step-3.mdx
@@ -428,21 +428,12 @@ curl -L -X POST -H "Content-Type: application/json" -d '{
       "method": "sidechain_getStatus",
       "params": [INSERT_EPOCH],
       "id": 1
-    }' https://rpc.testnet-02.midnight.network | jq
+    }' https://rpc.preview.midnight.network | jq
 ```
 
 Example output:
 
-```bash
-curl -L -X POST -H "Content-Type: application/json" -d '{
-      "jsonrpc": "2.0",
-      "method": "sidechain_getStatus",
-      "params": [],
-      "id": 1
-    }' https://rpc.testnet-02.midnight.network | jq
-  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
-                                 Dload  Upload   Total   Spent    Left  Speed
-100   297  100   193  100   104    395    213 --:--:-- --:--:-- --:--:--   608
+```json
 {
   "jsonrpc": "2.0",
   "result": {
@@ -469,7 +460,7 @@ curl -L -X POST -H "Content-Type: application/json" -d '{
       "method": "systemParameters_getAriadneParameters",
       "params": [697],
       "id": 1
-    }' https://rpc.testnet-02.midnight.network | jq
+    }' https://rpc.preview.midnight.network | jq
 ```
 
 Look for your registration under `"candidateRegistrations": {)`. Ensure you see your Partner-chain keys in the registration and  `"isValid": true`
@@ -516,7 +507,7 @@ curl -L -X POST -H "Content-Type: application/json" -d '{
       "method": "systemParameters_getAriadneParameters",
       "params": [INSERT_EPOCH],
       "id": 1
-    }' https://rpc.testnet-02.midnight.network | jq
+    }' https://rpc.preview.midnight.network | jq
 ```
 
 You should see your candidate is no longer on the list.
--- a/main_versioned_docs/version-0.0.0/develop/guides/wallet-dev-guide.mdx
+++ b/main_versioned_docs/version-0.0.0/develop/guides/wallet-dev-guide.mdx
@@ -131,10 +131,10 @@ import { WalletBuilder } from '@midnight-ntwrk/wallet';
 import { NetworkId } from '@midnight-ntwrk/zswap';
 
 const wallet = await WalletBuilder.build(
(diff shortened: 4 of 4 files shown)


docs/nodes/_run-a-validator/index.mdx
```diff
@@ -171,7 +171,7 @@ An existing Cardano relay node can be repurposed for `cardano-db-sync` if the op
 
 Midnight-node uses partner-chains CLI commands which require on an Ogmios connect. While you may run Ogmios locally, we provide a public endpoint for the service.
 
-* [ogmios.testnet-02.midnight.network](https://ogmios.testnet-02.midnight.network)
+* The former public Ogmios endpoint (`ogmios.testnet-02.midnight.network`) has been retired along with the `testnet-02` network and no longer resolves. Run Ogmios locally, and see [Networks and environments](/guides/networks-and-environments) for the current endpoints.
```
oduameh, line 174: The retained sentence above still promises a public endpoint, which now contradicts your new note. Folding the two together also avoids pointing readers at the networks guide for Ogmios, which that guide does not list. I verified the DNS side: the old name is NXDOMAIN and no preview Ogmios endpoint exists.

```suggestion
Midnight-node uses partner-chains CLI commands, which require an Ogmios connection. The former public endpoint (`ogmios.testnet-02.midnight.network`) was retired with the `testnet-02` network, so run Ogmios locally; the compose file in step 2 includes it.
```

docs/nodes/_run-a-validator/step-1.mdx
```diff
@@ -13,7 +13,7 @@ To embark on the path of becoming a Midnight validator, one must either be, or b
 
 | Cardano env. | Status | Midnight env.
 |----------|----------|----------|
-| `preview` | ✅ | `testnet-02` |
+| `preview` | ✅ | `preview` (the `testnet-02` network has been retired; see [Networks and environments](/guides/networks-and-environments)) |
```
oduameh, line 16: The mapping is right (I verified `system_chain` and the epoch alignment against Cardano preview), but an unqualified check mark oversells the current state: the live chain shows zero registered candidates receiving committee seats under the present D-parameter on both preview and preprod. A short hedge keeps it honest, and the parenthetical can lose the retirement note since the guide only needs it once.

```suggestion
| `preview` | ✅ | `preview` (committee seats for registered candidates depend on the current D-parameter) |
```

docs/nodes/_run-a-validator/step-3.mdx
```diff
@@ -428,21 +428,12 @@ curl -L -X POST -H "Content-Type: application/json" -d '{
       "method": "sidechain_getStatus",
       "params": [INSERT_EPOCH],
```
oduameh, line 429: `sidechain_getStatus` takes no parameters (verified live: identical result with `[]` and with an epoch), and your corrected example output below already shows the `[]` call. Note this line is also fixed in #1252, see the coordination note in the review body.

```suggestion
      "params": [],
```

docs/nodes/_run-a-validator/step-3.mdx
```diff
@@ -469,7 +460,7 @@ curl -L -X POST -H "Content-Type: application/json" -d '{
       "method": "systemParameters_getAriadneParameters",
       "params": [697],
```
oduameh, line 461: The literal `697` is a testnet-02-era value (the old example output shows mainchain epoch 695, plus 2) and it hangs indefinitely against `rpc.preview.midnight.network`; I tested it at 20s and 35s timeouts with zero bytes back, while the current mainchain epoch plus 2 answers in about 9 seconds. Restoring the placeholder keeps the command working forever. The retained example output above this block also still shows the testnet-02-era epochs (sidechain 239830, mainchain 695) and is worth refreshing from a live preview response while you are in the file.

```suggestion
      "params": [INSERT_EPOCH],
```

main_versioned_docs/version-0.0.0/develop/guides/wallet-dev-guide.mdx
```diff
@@ -131,10 +131,10 @@ import { WalletBuilder } from '@midnight-ntwrk/wallet';
 import { NetworkId } from '@midnight-ntwrk/zswap';
 
 const wallet = await WalletBuilder.build(
-  'https://indexer.testnet-02.midnight.network/api/v1/graphql', // Indexer URL
-  'wss://indexer.testnet-02.midnight.network/api/v1/graphql/ws', // Indexer WebSocket URL
+  'https://indexer.preview.midnight.network/api/v4/graphql', // Indexer URL
+  'wss://indexer.preview.midnight.network/api/v4/graphql/ws', // Indexer WebSocket URL
```
oduameh, line 135: Could you drop this file from the PR? Full reasoning in the review body: the legacy `WalletBuilder` stack this archived page documents cannot talk to the v4 indexer these URLs point at, so the edit makes the example look current while failing opaquely.

oduameh (changes requested):
Thanks for this, and for the careful PR description. I ran the whole thing through verification before reviewing, so the endpoint claims below are tested, not assumed.

What lands well: the four endpoint swaps are correct (`sidechain_getStatus` answers live on `rpc.preview` with real epoch data), the Ogmios retirement is real (NXDOMAIN, and no preview replacement exists), and the Example-output cleanup in step-3 fixes a genuine long-standing paste error.

Three things before this can merge:

1. **Please revert the `main_versioned_docs/version-0.0.0` wallet guide changes.** That snapshot is a frozen archive of the testnet-02 era, and the edit is refuted three ways: the legacy `@midnight-ntwrk/wallet` client syncs over the v1 `wallet(sessionId:)` subscription, which the v4 indexer no longer serves (live probe: `Unknown field "wallet" on type "Subscription"`); v4 rejects the legacy hex viewing keys (`cannot bech32m-decode viewing key`); and the NetworkId enum of that era (`Undeployed, DevNet, TestNet, MainNet`) cannot express the preview network at all. Twenty-three sibling files in the same snapshot still use testnet-02, so the archive would also become internally inconsistent. If you want to improve the archive, a single admonition at the top saying it documents the retired testnet-02 stack, linking to the networks guide, would be the right shape.

2. **Coordination with #1252**: the same four endpoint swaps in step-3 are already content-approved on that PR (open since 20 Aug, waiting only on a signed commit). To keep both contributions intact, could you rebase step-3 down to the parts #1252 does not have, the Example-output cleanup and the `[697]` fix, once it lands? I will coordinate timing so nothing gets wasted; if #1252 stalls, we will fold it back here.

3. **One gap that will strand readers**: step-3 still mounts `envs/testnet-02/pc-chain-config.json` from midnight-node-docker, whose presets today only ship `qanet` and `testnet-02` (its `.envrc` defaults to `CFG_PRESET=testnet-02`). After this PR, a reader registers against the retired chain parameters and then checks preview, where the registration never appears. Fixing that repo is out of scope here as your description says, but a `:::warning` admonition in step-3 noting the docker presets still target testnet-02 and a preview `pc-chain-config.json` is required would save readers a very confusing failure.

Two housekeeping notes: this section (`_run-a-validator`) is currently underscore-hidden from the live site, so these fixes matter for the repo and the eventual republish rather than for today’s readers, still worth doing; and the repo requires verified commit signatures, so the branch commit needs signing (squash with `git commit -S` and force-push) before merge.

Happy to re-review quickly once these are in.

Issue #1245, opened 2026-08-18
Follow-up: review findings on the two EffectStream guides


oduameh:
Both EffectStream guides were merged as-is by direction, ahead of the review findings being addressed. This issue tracks everything flagged during review of #1240 and #1241 so none of it is lost.

Pages:
- `docs/guides/index-state-with-effectstream.mdx` (from #1240)
- `docs/guides/build-cross-chain-dapp-with-effectstream.mdx` (from #1241)

Full detail, with line-anchored suggestions, is in the reviews on [#1240](https://github.com/midnightntwrk/midnight-docs/pull/1240) and [#1241](https://github.com/midnightntwrk/midnight-docs/pull/1241). Many findings there are one-click committable suggestions.

## Fix first

- [ ] **Broken link on both pages.** Each links `./install-bun-runtime-midnight`, which resolves to `/guides/install-bun-runtime-midnight`. That page sets `slug: /how-to/bun-runtime-midnight`, so the route does not exist. Confirmed by build; `onBrokenLinks` is `warn`, so the build passes and readers get the 404. Use `/how-to/bun-runtime-midnight`.
- [ ] **Root cause of the above.** `install-bun-runtime-midnight.mdx` has a custom slug its own filename does not match, so any page linking it relatively breaks the same way. Either add a `/guides/install-bun-runtime-midnight` redirect to `vercel.json`, or drop the custom slug.
- [ ] **`disclose()` described inaccurately, cross-chain guide.** "what a `disclose()` call publishes" contradicts [Security and best practices](https://docs.midnight.network/guides/security-best-practices#on-chain-visibility), which the same page links twice. `disclose()` does not publish; it clears the compiler's private-data check so a value may cross into a public position, and the ledger write is what makes it visible. Same wording repeats in that page's Additional resources list.
- [ ] **`compact update 0.31.0` has two problems, cross-chain guide.** The command switches the machine-wide default compiler (per its own `--help`), so following the prerequisite repoints every other Midnight project on the reader's machine. Use `--no-set-default`. Separately, `docs/tokens/unshielded-token.mdx` states 0.31.0 "could drop a range-check constraint from generated ZK circuits for certain casts, which 0.31.1 fixes", so the guide currently pins readers to a compiler the docs warn about.
- [ ] **Undocumented hard prerequisites, cross-chain guide.** Docker and `git` are never named, though the template brings up the Midnight node, indexer, proof server, and database as containers.

## Third-party placement

[...]

PR #1311, opened 2026-09-05
docs(guides): fix links, fences, style in EffectStream guides

tirthraj01:
## Summary
Fixes link, style, and prerequisite issues flagged in #1245 for the two EffectStream guides.

- `docs/guides/index-state-with-effectstream.mdx`
- `docs/guides/build-cross-chain-dapp-with-effectstream.mdx`

## Changes made
- Fixed Bun runtime link to absolute `/how-to/bun-runtime-midnight` on both pages (404 fix)
- Updated Compact toolchain installation instructions for `0.33.0-rc.2` prerelease
- Clarified prerequisites: Git and platform-specific Docker fallback
- Refined `disclose()` explanation to accurately describe on-chain ledger visibility
- Standardized code fences to `typescript` across both pages (repo convention)
- Sentence casing for Compact toolchain and consistent pronoun style (`you`)

Contributes to #1245


docs/guides/build-cross-chain-dapp-with-effectstream.mdx
```diff
@@ -20,9 +20,11 @@ Use this guide to build a DApp whose state lives on Midnight and an EVM chain at
 
 These apply to every procedure in this guide:
 
-- [Bun](https://bun.sh). See [Set up Bun for Midnight development](./install-bun-runtime-midnight).
-- [Foundry](https://www.getfoundry.sh/). `forge` compiles the Solidity artifacts, and the orchestrator checks for it on PATH before starting.
-- The Compact compiler, which the orchestrator also checks on PATH. The template pins `0.31.0`, so run `compact update 0.31.0`.
+- [Git](https://git-scm.com/) for cloning the template repository.
+- [Docker](https://www.docker.com/) (or Docker Desktop) running, used by the orchestrator to start containerized services including the Midnight node, proof server, indexer, and database.
+- [Bun](https://bun.sh). See [Set up Bun for Midnight development](./how-to/bun-runtime-midnight).
```
oduameh: Relative links resolve against the page URL, so this one lands on `/guides/how-to/bun-runtime-midnight`; the build's broken-link report lists it for this page. The absolute path works.

```suggestion
- [Bun](https://bun.sh). See [Set up Bun for Midnight development](/how-to/bun-runtime-midnight).
```

docs/guides/index-state-with-effectstream.mdx
```diff
@@ -20,11 +20,11 @@ The procedures target the `preview` network and a contract you need not deploy,
 
 These apply to every procedure in this guide:
 
-- [Bun](https://bun.sh). EffectStream runs TypeScript directly and uses Bun as its package manager. See [Set up Bun for Midnight development](./install-bun-runtime-midnight).
+- [Bun](https://bun.sh). EffectStream runs TypeScript directly and uses Bun as its package manager. See [Set up Bun for Midnight development](./how-to/bun-runtime-midnight).
```
oduameh: Same resolution problem as the cross-chain guide: this becomes `/guides/how-to/bun-runtime-midnight`. The absolute path is what the slug on the Bun page expects.

```suggestion
- [Bun](https://bun.sh). EffectStream runs TypeScript directly and uses Bun as its package manager. See [Set up Bun for Midnight development](/how-to/bun-runtime-midnight).
```

docs/guides/build-cross-chain-dapp-with-effectstream.mdx
```diff
@@ -20,9 +20,11 @@ Use this guide to build a DApp whose state lives on Midnight and an EVM chain at
...
+- [Foundry](https://www.getfoundry.sh/). `forge` compiles the Solidity artifacts, and the orchestrator checks for it on `PATH` before starting.
+- The Compact compiler, which the orchestrator also checks on `PATH`. The template pins `0.31.1`, so run `compact update --no-set-default 0.31.1`.
```
oduameh: The template's build script on `v-next` is `compact compile +0.33.0-rc.2 src/counter.compact src/managed`, so it picks the toolchain itself and the default never comes into play. That prerelease is not in the devtools index, so `compact update 0.33.0-rc.2` fails with "Couldn't find version" whether or not `--no-set-default` is there, and `compact compile +0.33.0-rc.2` does not download it on demand. What does work, and what EffectStream's CI Dockerfile does, is dropping the release zip into the devtools layout. I verified the wording below on Apple silicon: the version check prints `0.33.0` and the default toolchain stays untouched.

```suggestion
- The Compact toolchain, which the orchestrator checks on `PATH` before starting. The template's build script selects toolchain `0.33.0-rc.2` explicitly, and `compact update` cannot install prereleases, so download `compactc_v0.33.0-rc.2_<target>.zip` from the [Compact releases](https://github.com/LFDT-Minokawa/compact/releases/tag/compactc-v0.33.0-rc.2) and unzip it into `~/.compact/versions/0.33.0-rc.2/<target>/`, using the same `<target>` as the zip name, for example `aarch64-darwin`. Your default toolchain stays as it is. Confirm with `compact compile +0.33.0-rc.2 --version`, which prints `0.33.0`.
```

One thing worth knowing rather than fixing here: 0.33 targets ledger v9, which the 0.34.0 release note says is not on Mainnet yet. The template runs its own local node, so the guide still works, but it is no longer on the supported 0.31.x line.

docs/guides/build-cross-chain-dapp-with-effectstream.mdx
```diff
@@ -20,9 +20,11 @@ Use this guide to build a DApp whose state lives on Midnight and an EVM chain at
 
 These apply to every procedure in this guide:
 
-- [Bun](https://bun.sh). See [Set up Bun for Midnight development](./install-bun-runtime-midnight).
-- [Foundry](https://www.getfoundry.sh/). `forge` compiles the Solidity artifacts, and the orchestrator checks for it on PATH before starting.
-- The Compact compiler, which the orchestrator also checks on PATH. The template pins `0.31.0`, so run `compact update 0.31.0`.
+- [Git](https://git-scm.com/) for cloning the template repository.
+- [Docker](https://www.docker.com/) (or Docker Desktop) running, used by the orchestrator to start containerized services including the Midnight node, proof server, indexer, and database.
```
oduameh: The orchestrator does not start these as containers. `packages/contracts-midnight/package.json` launches the node, indexer, and proof server through the `@effectstream/npm-midnight-*` packages, which ship native binaries for `linux-amd64` and `macos-arm64`, and only the indexer and proof server have a Docker fallback for other platforms. There is no docker-compose anywhere in the repo, and the database step uses embedded PGlite by default.

```suggestion
- [Docker](https://www.docker.com/), only on platforms without native Midnight binaries. On Linux x86-64 and Apple silicon the orchestrator runs the Midnight node, indexer, and proof server as native binaries; on other platforms it falls back to Docker images for the indexer and proof server.
```

Since the node itself has no Docker fallback, you may prefer to state the two supported platforms outright and drop Docker from the list. Either reads fine to me.

docs/guides/build-cross-chain-dapp-with-effectstream.mdx
```diff
@@ -34,11 +36,11 @@ Splitting a DApp across chains usually means a bridge: a message-passing contrac
...
+The template splits an NFT along that line. An ERC-721 contract on the EVM chain owns transfers, because everyone needs to agree on who holds a token. A Compact circuit on Midnight takes private inputs and discloses one resulting property, because the values behind it stay private. The two sides share a key, `(contract_address, token_id)`, which you supply to both. The rollup joins on that key.
 
-One consequence shapes the code you write. EffectStream reads Midnight's public ledger, so what a `disclose()` call publishes is what your state machine can see. Designing the circuit designs the sync surface.
+One consequence shapes the code you write. EffectStream reads Midnight's public ledger, so values disclosed in circuits and committed to the ledger are what your state machine can see. Designing the circuit designs the sync surface.
```
oduameh: The meaning is right now, thank you. One word: "committed" has a specific sense in Midnight, where a commitment is the hiding primitive behind nullifiers and the `commitments` map in the sibling guide, so a reader can take this as the opposite of what you mean. "Writes to the ledger" says the same thing without the collision.

```suggestion
One consequence shapes the code you write. EffectStream reads Midnight's public ledger, so values a circuit discloses and writes to the ledger are what your state machine can see. Designing the circuit designs the sync surface.
```

oduameh (changes requested):
Thanks for picking this up, Tirth, and for going past the checklist to the `disclose()` and prerequisite items. The style fixes are all good and I would take them as they are. Three of the technical edits need another pass. Two of them trace back to the template itself: it changed its compiler pin on 25 August and its README has not caught up, which is where the mismatch comes from.

The link. The diff uses `./how-to/bun-runtime-midnight`, and Docusaurus resolves a relative path against the page URL, so it lands on `/guides/how-to/bun-runtime-midnight`. I built the branch and both pages appear in the broken-link report with exactly that resolution. The absolute path from your PR description is the right one; it just did not make it into the diff.

The compiler. On the template's default branch (`v-next`, there is no `main`) the build script now runs `compact compile +0.33.0-rc.2`. It was 0.31.0 before commit eb0542e9 and was never 0.31.1; the template README and Dockerfile still say 0.31.0, which is where the mismatch comes from. Because the script names the toolchain itself, `--no-set-default` makes no difference either way, and `compact update` cannot fetch that prerelease at all. I tried it with devtools 0.5.1 and got "Couldn't find version 0.33.0-rc.2"; EffectStream's own CI Dockerfile side-loads the zip for the same reason. The inline suggestion carries an install path I ran end to end on macOS.

Docker. The orchestrator runs the Midnight node, indexer, and proof server as native binaries on Linux x86-64 and Apple silicon, and only falls back to Docker images for the indexer and proof server on other platforms. There is no docker-compose in the repo. The suggestion says that instead of listing containers.

Two small ones. "committed to the ledger" reads as a cryptographic commitment in Midnight, so "writes to the ledger" is safer; that one is inline. The other is the fifth `ts` fence from the checklist, at line 184 of the indexing guide (` ```ts title="index.ts" `). It sits outside this diff, which is why I could not attach a suggestion, but it is a one-word change to `typescript`.

The failing Vercel check is the fork-authorization gate rather than a build failure; the build passes. Once the suggestions land, the PR description could drop "no content changes" and "verified links resolve", since the second commit does change content and the link needs the absolute path. Happy to re-review as soon as you push.

tirthraj01:
Hi @oduameh, thank you for the detailed review and testing! 
I have applied all your suggestions and pushed the updates:

- Updated links to absolute /how-to/bun-runtime-midnight paths
- Added the toolchain 0.33.0-rc.2 manual download and unzip instructions
- Clarified native binaries on Linux/Apple Silicon with Docker fallback
- Adjusted the disclose() wording to 'values a circuit discloses and writes to the ledger'
- Fixed the line 184 code fence to typescript in the indexing guide
- Updated the PR description as requested

Ready for re-review!

oduameh (approved):
Thanks Tirth, this all looks good now, and thanks for catching the fence outside the diff too. I merged it with the latest main locally and the build is clean, with no broken links on either page.

One thing on my side, though. The Docker line I suggested was wrong, since only the indexer and proof server have a Docker fallback and the node won't run at all outside Linux x86-64 and Apple silicon. I've pushed a small commit to your branch that names those two platforms instead, so there's nothing more you need to do here. Sorry about that, and thanks for sticking with this one.

PR #1314, opened 2026-09-08
docs: make the keeping-data-private Merkle examples compile and bind paths to the leaf

wbaxterh:
## Summary

Both Merkle membership examples in `docs/concepts/how-midnight-works/keeping-data-private.mdx` have two problems: they do not compile, and they accept a witness path that belongs to a different leaf. This fixes both. It supersedes #1225 and carries that work with @Jc-asastu credited as co-author.

**Compilation.** I extracted both `compact` blocks from the page verbatim and compiled them. Five separate errors, in three groups:

- The two syntax fixes from #1225: `merkleTreePathRoot` takes the path, not `path.value`, and `assert` needs the function form.
- @oduameh's three `disclose()` suggestions from the #1225 review, applied as written.
- Two disclosures that review did not reach: `items.insert(item)` and `authorizedCommitments.insert(pk)`. Both are exported-circuit parameters flowing into a ledger write, so the compiler requires `disclose()` even though they look public. With only #1225 and the review suggestions applied, the page still fails on these two.

**Binding.** `findItem` and `findAuthPath` are witnesses, and every Merkle path in the tree is public data. `checkRoot` alone therefore proves that *some* leaf is in the tree, not that *yours* is. In `check`, any valid path passes for any item. In `increment`, anyone who observed a valid path in a public transaction can replay it under a secret key that was never authorized, and the nullifier check does not catch it because that key's nullifier is fresh.

The fix is one assert per example, comparing `path.leaf` to the value the circuit computed. This is not a new rule: the security guide already calls it "the security-critical line" and shows this exact shape in [Restricting a circuit to a group](https://docs.midnight.network/guides/security-best-practices#restricting-a-circuit-to-a-group). The concept page where most people first meet Merkle membership omits it in both examples and does not link there, so this adds the asserts plus one sentence after each example pointing at the guide. An `assert` over private values discloses nothing, so no new disclosure is introduced.

**Verification.** Both blocks, extracted from the page as rendered, compile on compactc 0.31.1 (`compact` CLI 0.5.1): 4 circuits, prover and verifier keys emitted for `insert`, `check`, `addAuthority` and `increment`. Note for anyone repeating this: `compact compile` prints `Exception:` on failure but still exits 0, so check the output or the emitted artifacts rather than the exit code.

## Type of Change

- [x] Documentation update (content fixes, new pages)

## Related Links

- Documentation page(s) updated: `/concepts/how-midnight-works/keeping-data-private`
- Related guide: `/guides/security-best-practices#restricting-a-circuit-to-a-group`

## Issue Reference

Supersedes #1225.

## Checklist

- [x] Followed the [Midnight Style Guide](https://docs.midnight.network/contributing/style-guide).
- [x] Previewed changes locally (`yarn build`).
- [ ] Updated navigation metadata if needed (sidebar, menus). Not needed.

## Note on #1225

@Jc-asastu opened #1225 on Aug 11 with the two syntax fixes and has not been active on it since, including through @oduameh's review. I said on that thread five days ago that I would carry it in a follow-up with credit if they were busy. That is what this is: their two fixes are here, they are `Co-authored-by` on the commit, and I am happy for this to close in favour of #1225 instead if they would rather finish it themselves.


docs/concepts/how-midnight-works/keeping-data-private.mdx
```diff
@@ -222,3 +226,5 @@ circuit nullifier(sk: Bytes<32>): Bytes<32> {
   return persistentHash<Vector<2, Bytes<32>>>([pad(32, "nullifier-domain"), sk]);
 }
 ```
+
+The `authPath.leaf == publicKey(sk)` assert binds the path to the caller's key. Merkle paths are public tree data and `findAuthPath` is a witness, so without that line anyone who observed a valid path in a public transaction could replay it with an unauthorized secret key. See [Restricting a circuit to a group](../../guides/security-best-practices#restricting-a-circuit-to-a-group) in the security guide.
```
oduameh: The path is never part of the transaction, only the disclosed root is, and the leaves are public, so anyone can compute a valid path for an authorized key without observing anything. Saying so keeps the threat model precise, and it makes the nullifier point explicit.

```suggestion
The `authPath.leaf == publicKey(sk)` assert binds the path to the caller's key. The tree's leaves are public, so anyone can compute a valid path for an authorized key, and `findAuthPath` is a witness the prover controls. Without that line, a secret key that was never authorized could pass with someone else's path, and its fresh nullifier would not stop it. See [Restricting a circuit to a group](../../guides/security-best-practices#restricting-a-circuit-to-a-group) in the security guide.
```

oduameh (dismissed):
.

oduameh (approved):

PR #1345, opened 2026-09-16
Add node v1.0.2 release notes

MidnightCI:
Automated release notes copy for **node v1.0.2**.

Source: `https://github.com/midnightntwrk/midnight-node/releases/tag/node-1.0.2`

### What to fill in manually

- [ ] `summary`: replace the placeholder in `src/components/DynamicListNode.js`
- [ ] `details`: add 5-8 bullet strings describing user-facing changes
- [ ] `artifacts`: add NPM/Docker links
- [ ] Reformat `docs/relnotes/node/node-1-0-2.mdx` to match the existing version file template


oduameh:
Replaced the raw cumulative changelog the workflow copied in with the delta from 1.0.1, taken from the `changes/` fragments between the `node-1.0.1` and `node-1.0.2` tags: the tblock validation fix that stalled Preview at 128536 (three PRs), the two new config values with their defaults checked against `res/cfg/default.toml`, ledger 8.1.2, and the toolkit image security fixes. Runtime stays 1.0.0. DynamicList entry filled in.

Build check: full `yarn build` of main with this branch plus #1346, #1347, #1348 and #1349 merged succeeds; none of the new or changed release-note pages have broken links (the warnings in the log are the pre-existing 0.0.0 and blog ones).

oduameh:
I pushed a correction to this branch, and it is worth reading before merging.

The page as it stood described the block timestamp correction work: the `tblock_correction_offset` and `tblock_correction_disable_after` settings, the Preview stall, and the PRs behind them. That content was accurate against the release body as it read on 16 September, but the tag was re-cut on 18 September and none of that work is in the published tag.

Three checks:

- `node-1.0.1...node-1.0.2` is 15 commits across 16 files, with nothing under `pallets/` or `ledger/src/versions/` and no tblock commits.
- At the tag, neither `ledger/src/versions/block_context/post_ledger_8.rs` nor `node/src/cfg/midnight_cfg/mod.rs` contains `tblock_correction`.
- At the tag, `changes/node/changed/` has no tblock fragment. Those fragments now live only in `.changes_archive/node-1.0.2/` on `main`, and `release/node-1.0.2` points at the same commit as the tag.

So the page now documents what actually shipped: the ledger pin moving from 8.1.1 to 8.1.2 with its companion crates, the toolkit npm audit fixes from #2179, the toolkit image moving to Node.js 24 LTS along with the resolve-hook re-entrancy guard that change required, and refreshed apt pins in the local environment contract compiler image. The runtime is untouched, so there is no runtime upgrade. I also corrected the date to 18 September, which matches both the release and the Docker pushes.

The correction is a separate commit on purpose, so you can see exactly what came out and why.

One question for the node team before this merges: was dropping the tblock correction from the re-cut 1.0.2 intentional? It merged into the release branch in July and August and is archived on `main` under that version. If it was meant to ship, the release needs another look rather than the docs. If it is coming in a later version, this page is right as written.

oduameh:
@nstanford5 when you have a moment, could I get your eyes on this one? It is the node 1.0.2 notes, and it matters more than a normal release notes review because the tag was re-cut on 18 September and I had to pull content back out that the earlier draft had in it. My comment above lays out what changed and how I checked it.

There is also a question for your team about the block timestamp work, which I have asked on #1355 so it does not get lost when this merges. No rush on that one, but I would rather not merge the page until we know the answer.

nstanford5:
Blocked per the comment here: https://github.com/midnightntwrk/midnight-docs/issues/1355#issuecomment-5770161675

PR #1347, opened 2026-09-16
Add Midnight.js v4.1.0 release notes

oduameh:
## Summary
Midnight.js 4.1.0 (20 May) had two breaking changes and never got a page; we only have 4.1.1, which does not fold it in. The GitHub release body is a commit list, so this page is written from the curated notes in the repo (`docs/releases/v4.1.0`: release-notes.md, breaking-changes.md, migration-guide.md), in the format of the 4.1.1 page.

- Breaking: `StorageEncryption` is async with a `StorageEncryption.create()` factory; protocol imports go through `@midnight-ntwrk/midnight-js-protocol` subpaths, with the before/after table and the ESLint rule that reports the old form.
- Behaviour changes called out separately: PBKDF2 `verifyPassword` cost on cold reads, `assertDefined`/`assertUndefined` accepting falsy values, `decryptValue` failing closed.
- Features: Web Crypto storage encryption with the `CryptoBackend` abstraction and Noble fallback, injectable `levelFactory` for React Native, the protocol package, `DAppConnectorWalletAdapter` in testkit-js, `assertSafeName`/`assertSemVer`.
- Fixes: shielded coins from fallible-segment circuits routed to the correct offer, browser-bundleable `midnight-js-contracts`, path traversal blocked in the ZK providers, HTML responses rejected by `FetchZkConfigProvider`.
- DynamicList: 4.1.0 entry as SUPPORTED between 4.1.1 and 4.0.4. The 4.0.4 entry still had the `Summary of Release 4.0.4` placeholder; filled from its own details.

Matrix unchanged (4.1.1 stays LATEST).

## Type of Change
- [x] Documentation update (content fixes, new pages)
- [ ] Site improvement (UI/UX, infrastructure)
- [ ] New blog post
- [ ] Other (please specify)

## Related Links
- Documentation page(s) updated (if applicable): `docs/relnotes/midnight-js/midnight-js-4-1-0.mdx`, `src/components/DynamicListMidnightJS.js`
- Other relevant resources (if any): https://github.com/midnightntwrk/midnight-js/tree/main/docs/releases/v4.1.0

## Issue Reference
None

## Checklist
- [x] Followed the [Midnight Style Guide](https://docs.midnight.network/contributing/style-guide).
- [ ] Previewed changes locally (`npm run start` or `yarn start`).
- [ ] Updated navigation metadata if needed (sidebar, menus).

## Notes
[...]

PR #1346, opened 2026-09-16
Add Midnight Indexer 3.1.0, 4.0.1, 4.0.2, 4.1.0, 4.2.0, 4.2.1, 4.3.0 and 4.3.1 release notes

oduameh:
## Summary
Backfill of the indexer release notes we were missing. The release watch only compares the single newest stable tag with the newest DynamicList entry, so eight stable releases between 3.0.0 and 4.3.2 never got a page (I am fixing the watch separately). Each page is written from the GitHub release body for that tag, in the format of the existing 4.3.2 and 4.3.3 pages, with PR links kept.

- New pages: 3.1.0 (last v3 API release, Node 0.21), 4.0.1 and 4.0.2 (the 4.0.x maintenance line for Node 0.22.x; 4.0.2 carries the re-index warning), 4.1.0 (first Node 1.0 release: Zswap collapsed update, dust wallet sync, `shieldedNullifierTransactions`, `fee` field), 4.2.0, 4.2.1, 4.3.0 and 4.3.1.
- DynamicList: entries for all eight, in date order between 4.3.2 and 4.0.2 and between 4.0.0 and 3.0.0. The 4.0.1 and 4.0.2 entries existed as stubs with no `link`; they now point at their pages.
- Fixed the 2.1.4 entry, whose `link` went to the 2.1.3 page (known issue 2 in the release notes guide).

No change to the matrix: 4.3.3 stays the newest entry and LATEST. Statuses follow the existing pattern (SUPPORTED for the 4.x line, UNSUPPORTED for 3.1.0 next to 3.0.0).

## Type of Change
- [x] Documentation update (content fixes, new pages)
- [ ] Site improvement (UI/UX, infrastructure)
- [ ] New blog post
- [ ] Other (please specify)

## Related Links
- Documentation page(s) updated (if applicable): `docs/relnotes/midnight-indexer/midnight-indexer-{3-1-0,4-0-1,4-0-2,4-1-0,4-2-0,4-2-1,4-3-0,4-3-1}.mdx`, `src/components/DynamicListMidnightIndexer.js`
- Other relevant resources (if any): the watch fix that would have caught these is in the `ci/release-watch-holes` PR

## Issue Reference
None

## Checklist
- [x] Followed the [Midnight Style Guide](https://docs.midnight.network/contributing/style-guide).
- [ ] Previewed changes locally (`npm run start` or `yarn start`).
- [ ] Updated navigation metadata if needed (sidebar, menus).

## Notes
[...]

PR #1348, opened 2026-09-16
Add Compact toolchain 0.31.1 release notes and fill the Compact list summaries

oduameh:
## Summary
Toolchain 0.31.1 is the version the compatibility matrix points at, and it had no page: the release notes list jumps from 0.31.0 to 0.34.0. There are no notes for it anywhere upstream (the GitHub release body is empty and LFDT-Minokawa `doc/release-notes` has no 0.31.1 file), so the page is written from the CHANGELOG entry ("Fixed: code generation for certain casts") and the compiler diff between the `compactc-v0.31.0` and `compactc-v0.31.1` tags, which is a one-line change: the common-subexpression matching for unsigned downcasts ignored the `safe` flag, so a guarded and an unguarded downcast of the same value could be merged. Language 0.23.0 and runtime 0.16.0 are unchanged, and the page says so.

While in `DynamicListCompact.js`:
- 0.31.1 entry added as SUPPORTED with `compactVersion` 0.23.0 (0.34.0 stays LATEST).
- The 0.34.0, 0.31.0 and 0.30.0 entries were still the workflow stubs (`Summary of Release x`, empty details and artifacts). Filled from their pages, including the note on 0.34.0 that it targets ledger 9 and current deployments should stay on 0.31.

## Type of Change
- [x] Documentation update (content fixes, new pages)
- [ ] Site improvement (UI/UX, infrastructure)
- [ ] New blog post
- [ ] Other (please specify)

## Related Links
- Documentation page(s) updated (if applicable): `docs/relnotes/compact/toolchain-0.31.1.mdx`, `src/components/DynamicListCompact.js`
- Other relevant resources (if any): https://github.com/LFDT-Minokawa/compact/blob/compactc-v0.31.1/CHANGELOG.md

## Issue Reference
None

## Checklist
- [x] Followed the [Midnight Style Guide](https://docs.midnight.network/contributing/style-guide).
- [ ] Previewed changes locally (`npm run start` or `yarn start`).
- [ ] Updated navigation metadata if needed (sidebar, menus).

## Notes
The new page follows the format of the synced toolchain pages (no frontmatter, `# Compact toolchain x` heading), since the other files in `docs/relnotes/compact/` come from the compact repo through `release-notes.yml`. I have left 0.31.0 as UNSUPPORTED, which is what the workflow set when 0.34.0 came in; shout if you would rather it read SUPPORTED alongside 0.31.1. A full `yarn build` with this branch and the other release-note branches on main is recorded in the PR conversation once it finishes.

PR #1349, opened 2026-09-16
ci(release-watch): report holes behind the latest release and classify prereleases by tag, not by the GitHub flag

oduameh:
## Summary
Follow-up to #1290. The daily watch missed the release-note gaps I backfilled today, for two reasons:

1. It only compared the single highest stable tag with the newest DynamicList entry. Indexer 4.1.0 through 4.3.1 sat between documented 4.0.2 and 4.3.2 and never registered, because 4.3.3 was documented.
2. It trusted the GitHub prerelease flag, and the product repos do not set it consistently: `node-1.0.2` is flagged as a prerelease while the matrix runs it on mainnet, and `node-2.1.0-rc.1` is flagged as stable.

Changes to `scripts/check-release-gaps.mjs`:
- Releases are classified by tag shape only: clean `x.y.z` is stable, an `-rc.N`/`-beta.N`/`-alpha.N` suffix is a prerelease. Drafts are still skipped. The tag regexes now capture the suffix so prereleases can be reported.
- Every stable release with no DynamicList entry is a gap, as long as it is newer than the oldest entry the list still marks LATEST or SUPPORTED. That window keeps the script from opening issues for history nobody will document.
- One issue per missing version, same title format as before, so the "close to silence" behaviour is unchanged. The issue body now says to use the sync workflow only when the version is the newest, since the workflow always marks the new entry LATEST.
- The summary table gains a "Newer prerelease" column (information only, no issue), which is the early warning for the release-notes handoff: today it would show node 2.1.0-rc.1, ledger 10.1.0.0-alpha.1, indexer 4.4.0-rc.3, Midnight.js 5.0.0-beta.8, DApp Connector 4.1.0-beta.1, Compact.js 2.5.5-rc.8 and wallet SDK 2.0.0-beta.3.
- The ledger regex accepts the four-part `ledger-9.1.0.0` form.

## Type of Change
- [ ] Documentation update (content fixes, new pages)
- [x] Site improvement (UI/UX, infrastructure)
- [ ] New blog post
- [ ] Other (please specify)

## Related Links
- Documentation page(s) updated (if applicable): none
- Other relevant resources (if any): #1290 (original watch), #1345, #1346, #1347 and the Compact 0.31.1 PR (the backfill this would have caught)

## Issue Reference
None

## Checklist
- [x] Followed the [Midnight Style Guide](https://docs.midnight.network/contributing/style-guide).
- [ ] Previewed changes locally (`npm run start` or `yarn start`).
- [ ] Updated navigation metadata if needed (sidebar, menus).

## Notes
Tested offline with a stubbed GitHub API fed the real tag lists from the nine repos. Against current main it reports node 1.0.2; indexer 4.1.0, 4.2.0, 4.2.1, 4.3.0, 4.3.1 and 4.3.7; Midnight.js 4.0.3 and 4.1.0; Compact.js 2.5.2. Two of those are new to me: indexer `v4.3.7` (18 Aug, clean tag, flagged prerelease upstream, not in the indexer CHANGELOG) and `compact-js-v2.5.2` (tag exists, no page). Both can be closed as won't-document once the issues appear, which silences them for good. `node --check` passes; no dry run against the live API from my side, so the first scheduled run is the real test.


oduameh:
Once this merges, the first scheduled run will open issues for every hole it finds on main at that point. If #1345 to #1348 land first, that is down to indexer 4.3.7, Midnight.js 4.0.3 and Compact.js 2.5.2, which I suggest closing as won't-document (the 4.0.4 page already says 4.0.3 was not published, 4.3.7 is not in the indexer CHANGELOG, and 2.5.2 was superseded by 2.5.3 within a month).

